PR #9, opened 2016-02-11
Use service exception

mpharoah-d2l:
Finally got around to this:
Made the auth service token provisioning client throw an exception deriving from `ServiceException`. Once the LMS is updated to pull in the newest LOReS and Auth client packages, this should make identifying auth service connection errors easier.


mpharoah-d2l:
!set-reviewers @dylanjacobsd2l @j3parker

mpharoah-d2l:
(This is a low priority PR. It does not need to be merged this sprint.)

j3parker:
Opps I'm sorry, I missed this and made a conflict headache for you :(

j3parker:
Thanks a lot, fixing this up would be great. CC @omsmith

src/D2L.Security.OAuth2/D2L.Security.OAuth2.csproj
```diff
@@ -39,6 +39,9 @@
     <DebugSymbols>true</DebugSymbols>
   </PropertyGroup>
   <ItemGroup>
+    <Reference Include="D2L.Services.Core.Exceptions">
+      <HintPath>..\..\packages\D2L.Services.Core.Exceptions.1.0.0\lib\net451\D2L.Services.Core.Exceptions.dll</HintPath>
```
j3parker: - [ ] Get this dependency open-sourced and on public NuGet

src/D2L.Security.OAuth2/Provisioning/AuthServiceException.cs
```diff
@@ -0,0 +1,33 @@
...
+
+namespace D2L.Security.OAuth2.Provisioning {
+	
+	/// <summary>
+	/// Represents an error that occured when attempting to contact the Auth
+	/// Service or an error caused by the service's response
+	/// </summary>
+	public sealed class AuthServiceException : ServiceException {
+		
+		/// <inheritdoc/>
```
j3parker: Nice trick!

src/D2L.Security.OAuth2/Provisioning/Default/AuthServiceClient.cs
```diff
@@ -45,17 +46,60 @@ Uri authEndpoint
...
+		/// The auth service could not be reached, or it did not respond with
+		/// a status code indicating success.
+		/// </exception>
 		async Task<IAccessToken> IAuthServiceClient.ProvisionAccessTokenAsync(
 			string assertion,
 			IEnumerable<Scope> scopes
 		) {
 			string requestBody = BuildFormContents( assertion, scopes );
-			using( HttpResponseMessage response = await MakeRequest( requestBody ).SafeAsync() ) {
-				response.EnsureSuccessStatusCode();
```
j3parker: Hehe evil, glad to see it gone.

src/D2L.Security.OAuth2/Provisioning/Default/AuthServiceClient.cs
```diff
@@ -45,17 +46,60 @@ Uri authEndpoint
...
+				
+				if( !response.IsSuccessStatusCode ) {
+					throw new AuthServiceException(
+						errorType: ServiceErrorType.ErrorResponse,
+						serviceStatusCode: response.StatusCode,
+						message: string.Concat(
+							"The Auth Service did not respond with a status code indicating success. ",
+							"(Received ", ((int)response.StatusCode).ToString(), " ",
+							response.ReasonPhrase, ")"
+						)
```
j3parker: In this and the one below, could we bundle the auth service errors into this error? (If you can think of reasons not to that's cool.)

It'd be nice to do that because people get these errors a lot and having all the details right away would help.

We could maybe try to see if it's a std OAuth2 error JSON but probably cool to proxy any errors?

src/D2L.Security.OAuth2/Provisioning/Default/AuthServiceClient.cs
```diff
@@ -45,17 +46,60 @@ Uri authEndpoint
...
+				}
+				
+				if( !response.IsSuccessStatusCode ) {
+					throw new AuthServiceException(
+						errorType: ServiceErrorType.ErrorResponse,
+						serviceStatusCode: response.StatusCode,
+						message: string.Concat(
+							"The Auth Service did not respond with a status code indicating success. ",
+							"(Received ", ((int)response.StatusCode).ToString(), " ",
+							response.ReasonPhrase, ")"
```
omsmith: This still isn't very useful, just a different version of the same old message.

I'd like to see an attempt for parse the (well-formed) error body if possible, and only fall back to this point if that didn't work.

@j3parker I had that in my PR, might it be accessible anywhere still (other than one of my computers)?

src/D2L.Security.OAuth2/Provisioning/Default/AuthServiceClient.cs
```diff
@@ -45,17 +46,60 @@ Uri authEndpoint
...
+				} catch( Exception exception ) {
+					throw new AuthServiceException(
+						errorType: ServiceErrorType.ClientError,
+						message: "An error occurred in parsing the response from the Auth Service.",
+						innerException: exception
+					);
+				}
+			} finally {
+				if( response != null ) {
+					response.Dispose();
```
j3parker: Not sure if we have extensions here, but this is `.SafeDispose` in D2L land

src/D2L.Security.OAuth2/Provisioning/Default/AuthServiceClient.cs
j3parker: Agreed with this, it'd be cool (I'd merge this PR without it if you want to do it in a follow-up; I'm indifferent.)

The old PR is here: https://git.dev.d2l/projects/ZAR/repos/d2l.security.oauth2/pull-requests/74/overview

j3parker:
Looks good! Thanks for picking this one up.

src/D2L.Security.OAuth2/Provisioning/Default/AuthServiceClient.cs
omsmith: Here's (most of the relevant) bits of what I had, for reference. Think it's reasonable / could do something similar @mpharoah-d2l ?

```c#
using System.IO;
using System.Runtime.Serialization;
using System.Runtime.Serialization.Json;

namespace D2L.Security.OAuth2.Provisioning.Default {
	[DataContract]
	internal sealed class AuthServiceErrorBody {

		[DataMember]
		public string error { get; set; }

		[DataMember]
		public string error_description { get; set; }

		internal static bool TryParse( Stream bodyStream, out AuthServiceErrorBody result ) {
			var serializer = new DataContractJsonSerializer( typeof( AuthServiceErrorBody ) );

			try {
				result = (AuthServiceErrorBody)serializer.ReadObject( bodyStream );
				return true;
			} catch {
				result = null;
				return false;
			}
		}
	}
}
```

```c#
internal static async Task<AuthServiceProvisioningException> MakeFromResponse( HttpResponseMessage response ) {
			string message;

			try {
				using( var bodyStream = await response.Content.ReadAsStreamAsync().SafeAsync() ) {
					AuthServiceErrorBody errorBody;
					if( AuthServiceErrorBody.TryParse( bodyStream, out errorBody ) ) {
						message = string.Format( CultureInfo.InvariantCulture, "{0}: {1}", errorBody.error, errorBody.error_description );
					} else {
						bodyStream.Seek( 0, SeekOrigin.Begin );
						using( var reader = new StreamReader( bodyStream ) ) {
							message = string.Format( CultureInfo.InvariantCulture, "{0}: {1}", response.ReasonPhrase, reader.ReadToEnd() );
						}
					}
				}
			} catch( Exception ) {
				message = response.ReasonPhrase;
			}

			return new AuthServiceProvisioningException(
				errorType: ServiceErrorType.ErrorResponse,
				proposedStatusCode: HttpStatusCode.InternalServerError,
				message: string.Format( CultureInfo.InvariantCulture, MESSAGE_FORMAT, message ),
				serviceStatusCode: response.StatusCode
			);
		}
```

src/D2L.Security.OAuth2/Provisioning/Default/AuthServiceClient.cs
mpharoah-d2l: We're not including the D2L.Core package, so we don't have `SafeDispose()`

j3parker: There is `D2L.Services.Core.Extensions` but @brian-pearson was a bit unsure of having a library for just that. Daryl also had some thoughts on how to organize these libraries as well so it's probably best to avoid a premature dependency for this...

src/D2L.Security.OAuth2/Provisioning/Default/AuthServiceClient.cs
mpharoah-d2l: I can check for a json object of that format and bring the error title and description into the exception message (which will be placed in the `detail` field of the resulting json error response in the LMS that the API exception handler generates).

Alternatively, I could actually load the `error` and `error_description` into separate properties on the `AuthServiceException`, then update the API exception handler in the LMS to respond with a slightly different error type that includes these fields if the `ServiceException` is an `AuthServiceException`

j3parker: Interesting idea... `AuthServiceException` is pretty important so it doesn't seem 100% wrong to have its own handler. I do get a bit worried about an explosion in the LE but I also like the type-y nature (rather than just jamming more shit into a string.)

I'll defer to you to figure out what's best.

mpharoah-d2l: At some point, I'd actually like to improve the `ServiceException` base class to take in a request and (if available) response object, so we can get all information about what API call failed and how the service responded.

But for now, unless there are any objections, I think I'll just put the data in a single string and use the normal ServiceException. On leap week, I'll come back to this and think about making auth service errors their own error type.

mpharoah-d2l:
Are there any security concerns with relaying all or part of the auth service's response when it can't be serialized to an access token or error object? (So the LMS makes a call to the auth service, doesn't understand the response, and relays the response in it's error message to the user)

src/D2L.Security.OAuth2/Provisioning/Default/AuthServiceClient.cs
j3parker: Oh that's a slick idea.

SGTM.

j3parker:
Nope!

src/D2L.Security.OAuth2/Provisioning/Default/AuthServiceClient.cs
```diff
@@ -45,17 +57,72 @@ Uri authEndpoint
...
+				}
+				
+				throw new AuthServiceException(
+					errorType: ServiceErrorType.ClientError,
+					message: string.IsNullOrEmpty( responseJson ) ?
+						EMPTY_RESPONSE_ERROR_MESSAGE :
+						string.Concat( INVALID_JSON_ERROR_MESSAGE_PREFIX, responseJson )
+				);
+			} finally {
+				if( response != null ) {
```
j3parker: `.SafeDispose()` should now be available if you rebase: https://github.com/Brightspace/D2L.Services.Core.Extensions/blob/master/src/D2L.Services.Core.Extensions/System.IDisposable.Extensions.cs#L8

src/D2L.Security.OAuth2/Properties/AssemblyInfo.cs
```diff
@@ -32,9 +32,9 @@
 // You can specify all the values or you can default the Build and Revision Numbers
 // by using the '*' as shown below:
 // [assembly: AssemblyVersion("1.0.*")]
-[assembly: AssemblyVersion( "4.4.3.0" )]
-[assembly: AssemblyFileVersion( "4.4.3.0" )]
-[assembly: AssemblyInformationalVersion( "4.4.3.0" )]
+[assembly: AssemblyVersion( "4.5.0.0" )]
+[assembly: AssemblyFileVersion( "4.5.0.0" )]
+[assembly: AssemblyInformationalVersion( "4.5.0.0" )]
```
j3parker: It's preferable to not version bump within the PR because there are other changes going around... I'm going to bundle up a bunch of stuff and do a 5.0.0 shortly so could you remove this part?

mpharoah-d2l:
Pulled in the latest changes, refactored code to work with the latest changes, and made the last 2 changes you mentioned.

Re-ran all unit and integration tests (all passed), but I haven't done manual testing yet.

mpharoah-d2l:
Looks like the AppVeyor build is failing because I'm referencing version 1.0.0 of the D2L.Services.Core.Exceptions NuGet package, which is present in our internal D2L NuGet repository, but the public NuGet repo uses version 1.0.0.27289. The code should be identical, but the LMS and other services may need to be updated to use the same package version.

src/D2L.Security.OAuth2/Provisioning/Default/AuthServiceClient.cs
```diff
@@ -92,9 +165,18 @@ IEnumerable<Scope> scopes
 			var result = builder.ToString();
 			return result;
 		}
-
+		
 		private sealed class GrantResponse {
-			public string access_token { get; set; }
+			[JsonProperty( PropertyName = "access_token", Required = Required.Always )]
```
j3parker: +1 to using the good stuff now that we are forced to use Newtonsoft

test/D2L.Security.OAuth2.UnitTests/Provisioning/AuthServiceClientTests.cs
```diff
@@ -0,0 +1,165 @@
...
+using System;
+using System.Net;
+using System.Net.Http;
+using System.Threading.Tasks;
+
+namespace D2L.Security.OAuth2.Provisioning {
+	
+	[TestFixture]
+	[Category( "Unit" )]
+	internal sealed class AuthServiceClientTests {
```
j3parker: Nice tests

j3parker:
Ok, let me know when you want me to click merge.

mpharoah-d2l:
I'm good with this being merged now.

PR #29, opened 2016-02-14
Switch ID2LPrincipal.UserId to return a long and add ID2LPrincipal.ActualUserId

j3parker:


omsmith:
Reason for the switch to `long`?

j3parker:
Everyone has to do `long.Parse` anyway

PR #33, opened 2016-02-16
Uping major version of test framework version because the package no …

mtseD2L:
…longer includes D2L.Security.OAuth2.dll


mtseD2L:
@j3parker

j3parker:
See #23

PR #35, opened 2016-02-19
webapi: hintpath for local oauth2. catch up with oauth2 changes

omsmith:

PR #36, opened 2016-02-19
Dnxcore

omsmith:
- [x] Start tracking disposables again
- [x] Dispose of ECDsa
- [x] Dispose of RSA (waiting on nuget release)
- [ ] Investigate using `JsonWebKey`/`JsonWebKeySet` from IdentityModel


src/D2L.Security.OAuth2/Keys/Default/D2LSecurityToken.Disposal.cs
```diff
@@ -31,13 +31,6 @@ partial class D2LSecurityToken : IDisposable {
 					if( ShouldDispose() ) {
 						m_disposed = true;
 
-						foreach( var key in m_key.Values ) {
```
omsmith: Will have to look into this, given the `SecurityKey` themselves still aren't `IDisposable`

src/D2L.Security.OAuth2/Keys/Default/D2LSecurityToken.cs
```diff
@@ -31,8 +31,12 @@ internal sealed partial class D2LSecurityToken : SecurityToken {
 			m_validFrom = validFrom;
 			m_validTo = validTo;
 
-			m_key = new ThreadLocal<Tuple<AsymmetricSecurityKey, IDisposable>>(
-				valueFactory: keyFactory,
+			m_key = new ThreadLocal<AsymmetricSecurityKey>(
+				valueFactory: () => {
+					var key = keyFactory();
+					key.KeyId = id.ToString();
```
omsmith: Much nicer than `NamedKeySecurityKeyIdentifier` :+1:

src/D2L.Security.OAuth2/Keys/Default/D2LSecurityToken.cs
```diff
@@ -100,13 +97,25 @@ internal sealed partial class D2LSecurityToken : SecurityToken {
 			var key = GetKey();
 
 			if( key is RsaSecurityKey ) {
-				var csp = GetAsymmetricAlgorithm() as RSACryptoServiceProvider;
-				RSAParameters p = csp.ExportParameters( includePrivateParameters );
+				var rsaKey = key as RsaSecurityKey;
+
+				if( includePrivateParameters && !rsaKey.HasPrivateKey ) {
+					throw new Exception();
```
omsmith: Need a better `Exception`

src/D2L.Security.OAuth2/Keys/Default/EcDsaJsonWebKey.cs
```diff
@@ -91,9 +82,9 @@ string y
 				validFrom: DateTime.UtcNow,
 				validTo: ExpiresAt ?? DateTime.UtcNow + Constants.REMOTE_KEY_MAX_LIFETIME,
 				keyFactory: () => {
-					var cng = BuildEcDsaCng();
-					var key = new EcDsaSecurityKey( cng );
-					return new Tuple<AsymmetricSecurityKey, IDisposable>( key, cng );
+					var publicBlob = ECCPublicKeyBlobFormatter.Instance.BuildECCPublicBlob( this );
+					var key = new ECDsaSecurityKey( publicBlob, CngKeyBlobFormat.EccPublicBlob );
```
omsmith: Taking the blob directly is premium :+1:

src/D2L.Security.OAuth2/Keys/Default/EcDsaSecurityKey.cs
```diff
@@ -1,157 +0,0 @@
-using System;
-using System.IdentityModel.Tokens;
-using System.Security.Cryptography;
-
-namespace D2L.Security.OAuth2.Keys.Default {
-	internal sealed partial class EcDsaSecurityKey : AsymmetricSecurityKey {
```
omsmith: I was so pleased when I saw `ECDsaSecurityKey` in my autocomplete

src/D2L.Security.OAuth2/Keys/Default/EcDsaSecurityKey.cs
```diff
@@ -1,157 +0,0 @@
...
-					return false;
-				default:
-					throw new Exception( string.Format( "Unsupported algorithm '{0}", algorithm ) );
-			}
-		}
-
-		public override int KeySize {
-			get {
-				// This should be m_ECDsa.KeySize, but that is small, and microsoft complains
-				return IDENTITY_MODEL_MIN_KEY_SIZE;
```
omsmith: They fixed this hack for their stuff :+1:

src/D2L.Security.OAuth2/Keys/Default/RsaJsonWebKey.cs
```diff
@@ -77,10 +77,7 @@ string e
 				validFrom: DateTime.UtcNow,
 				validTo: ExpiresAt ?? DateTime.UtcNow + Constants.REMOTE_KEY_MAX_LIFETIME,
 				keyFactory: () => {
-					var rsa = new RSACryptoServiceProvider() { PersistKeyInCsp = false };
-					rsa.ImportParameters( m_parameters );
-					var key = new RsaSecurityKey( rsa );
-					return new Tuple<AsymmetricSecurityKey, IDisposable>( key, rsa );
+					return new RsaSecurityKey( m_parameters );
```
omsmith: Much better! :+1:

test/D2L.Security.OAuth2.UnitTests/Provisioning/Default/AccessTokenProviderTests.cs
```diff
@@ -62,7 +63,7 @@ await m_accessTokenProvider
 			var publicKeys = ( await m_publicKeyDataProvider.GetAllAsync().SafeAsync() ).ToList();
 
 			string expectedKeyId = publicKeys.First().Id.ToString();
-			string actualKeyId = m_actualAssertion.Header.SigningKeyIdentifier[ 0 ].Id;
+			string actualKeyId = m_actualAssertion.Header.Kid;
```
omsmith: Once again, much less obtuse. ++ :+1:

omsmith:
This is on top of #35

src/D2L.Security.OAuth2/Keys/Default/EcDsaSecurityKey.cs
j3parker: :joy:

j3parker:
I love this

omsmith:
https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/a2f781da4d310ad1b7c7b31d8fafa0d394ec7f91/src/Microsoft.IdentityModel.Tokens/AsymmetricSignatureProvider.cs#L454

So this does disposing of things. Need to figure out how/when this happens.

omsmith:
So `JwtSecurityHandler` gets it when `WriteToken` is called, but doesn't dispose of it or anything (`Dispose` is `protected`, so it wouldn't be able to anyway).

Seems we'll still need to track the `Disposables` ourselves, which in `KDev` is a possibility for `RSASecurityKey` (https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/KDev/src/Microsoft.IdentityModel.Tokens/RsaSecurityKey.cs#L48), but not in the latest nuget.

omsmith:
`JsonWebKey`/`JsonWebKeySet` is a thing in `IdentityModel.Tokens` now, https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/KDev/src/Microsoft.IdentityModel.Tokens/JsonWebKeySet.cs

omsmith:
So, as `JsonWebKey`/`JsonWebKeySet` are set up, we could use them for parsing.

`JsonWebKey` implements `SecurityKey`, and the `AsymmetricSignatureProvider` will do the correct thing with it, but it would have to do the correct thing multiple times, and no way for us to cache it. Sad.

src/D2L.Security.OAuth2/Keys/Default/EcDsaJsonWebKey.cs
omsmith: They undid this, since Cng isn't cross-platform, so they're trying to remove dependence on it

src/D2L.Security.OAuth2/Keys/Default/D2LSecurityToken.cs
```diff
@@ -100,13 +97,32 @@ internal sealed partial class D2LSecurityToken : SecurityToken {
...
+					publicParameters.Modulus = parameters.Modulus;
+					publicParameters.Exponent = parameters.Exponent;
+					parameters = publicParameters;
+				}
+
+				return new RsaJsonWebKey( KeyId, ValidTo, parameters );
+			} else if( key is ECDsaSecurityKey && !includePrivateParameters ) {
+				var ecDsaKey = key as ECDsaSecurityKey;
+				var ecDsaCng = ecDsaKey.ECDsa as ECDsaCng;
+				var cng = ecDsaCng?.Key;
```
omsmith: This area here is not cross-platform. It's as cross-platform as it was previously, but still not.

We might consider if we can have a `PrivateKeyProviders` output `JsonWebKey` or something, meaning it's already in place for us there, and then we can call `GetSignatureProvider` on that in order to get the `IDisposable` key things for usage.

Issue #34, opened 2016-02-17
Switch to LP-compatible Newtonsoft


j3parker:
I guessed the version number incorrectly :(

omsmith:
Alternatively, update LP and friends.

Gogo Newtonsoft 8.0.2!

j3parker:
Hm. That's a good point.
I'll ask around. I know some folks that were involved in the last one.

j3parker:
Deeming "close enough", but upgrading our Newtonsofts in LP etc. is still on my list

PR #38, opened 2016-03-03
Switch Newtonsoft to 6.0.6

j3parker:
FYI @mtseD2L 

Does it make sense to be this stupid paranoid? Or should we just do nothing?

I want to revisit all of this ASAP and understand things better but I want to get through the backlog of breaking changes in this next version more (sigh.)

RE: #34


omsmith:
No need between these two versions since they share the same `AssemblyVersion`

https://github.com/JamesNK/Newtonsoft.Json/blob/1497343173a181d678b4c9bbf60250a12f783f1c/Src/Newtonsoft.Json/Properties/AssemblyInfo.cs#L94
https://github.com/JamesNK/Newtonsoft.Json/blob/70d3e9327d5dcd9a1770b3aed5437d6aa0b15802/Src/Newtonsoft.Json/Properties/AssemblyInfo.cs#L93

j3parker:
Awesome thanks. I need to learn some stuffs.

PR #41, opened 2016-03-03
Fixing typo in readme.

mtseD2L:


j3parker:
Hey, thanks for the fix but as this is so out of date I just deleted it.

Originally we had this documentation within VS but at one point we decided to get rid of a bunch of it because there was a lot of churn at the time. This file got removed from VS but the deleted file must not have got committed (the rest did evidently.)

PR #44, opened 2016-03-30
auth service mock supports IDisposable to release underlying mock htt…

brian-pearson:
…p server


brian-pearson:
I don't really need IDisposable, but it seems like the friendly thing

What I really need to is to get the TestFramework guy published to nuget.   Last version we published was 1.3 to our local nuget server.   I presume we'd want to publish 2.0 (or whatever) version to nuget.org?   (not sure how to do that)

The other catch is that TestFramework depends on D2L.Security.OAuth2.   Hoping that TestFramework can depend on 4.4 and up (ie, not require 5.0) of D2L.Security.OAuth2.   This is because the latest and greatest of D2L.Services.Core is not compatible with 5.0 (ID2LPrincipal.UserId moved from being a string to a long). 

The other option is that we release a new version of D2L.Services.Core

omsmith:
:+1:

PR #45, opened 2016-04-21
package: uniform version numbers

### LINKED ISSUE #31: Consider a uniform version number for this suite of packages
This repo contains these NuGet packages:

    D2L.Security.OAuth2
    D2L.Security.OAuth2.WebApi
    D2L.Security.OAuth2.TestFramework

They are currently versioned independently.

I think it'd make sense for them to be versioned in lock-step because these libraries are highly coupled (by design - the second ones are extensions) so it is confusing to keep track of compatibility versions of each currently.

It would also have some nice advantages for development: you can put all csproj into the same sln and use solution references instead of bringing them in as NuGet packages (what `D2L.Security.OAuth2.WebApi` has always had to do with `D2L.Security.OAuth2`.) Right now the feedback loop from change in core -> changes in WebApi is really annoying.

I think this would be a good general approach for the "core library + extensions" pattern.

Thoughts? @omsmith @brian-pearson @mpharoah-d2l 

P.S.

The WebApi repo was merged into this one to make dev easier.

It matches the style of what MS is doing with .NET Core stuff (e.g. EntityFramework is many packages but one repo.)

AppVeyor CI also handles multiple-package case just swell.

omsmith:
Fixes #31

### DIFF
--- a/src/D2L.Security.OAuth2.TestFramework/Properties/AssemblyInfo.cs
+++ b/src/D2L.Security.OAuth2.TestFramework/Properties/AssemblyInfo.cs
@@ -32,8 +32,8 @@ using System.Runtime.InteropServices;
 // You can specify all the values or you can default the Build and Revision Numbers 
 // by using the '*' as shown below:
 // [assembly: AssemblyVersion("1.0.*")]
-[assembly: AssemblyVersion("2.0.0.0")]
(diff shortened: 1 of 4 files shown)


omsmith:
(Don't currently have a TestFramework built against 5.0.0)

j3parker:
:shipit:

PR #48, opened 2016-05-02
Use project references internally only, not nuget packages

j3parker:
Continuing on #31 

This moves all references to D2L.Security.OAuth2 to project references within this solution.

Nuget has an option IncludeReferencedProjects which looks like basically what we want ( https://docs.nuget.org/consume/command-line-reference )

Before Daryl pointed that out for me I came up with this. Turns out that AppVeyor doesn't "support" that yet and one of the workarounds in the issue was to do what I'm doing here. http://help.appveyor.com/discussions/suggestions/449-add-includereferencedprojects-to-nuget-pack-command

I manually built the WebApi package and verified that it only had the WebApi DLL and that its dependency list in the packed/compiled nuspec was sensible.


omsmith:
Post-merge :shipit:

j3parker:
:shipit:ed

PR #51, opened 2016-05-12
webapi: specify authorization attribute run order

omsmith:
I don't know that this is a particularly good way, but it appears to work and should be BC

### DIFF
--- a/src/D2L.Security.OAuth2.WebApi/Authorization/OAuth2AuthorizeAttribute.cs
+++ b/src/D2L.Security.OAuth2.WebApi/Authorization/OAuth2AuthorizeAttribute.cs
@@ -1,0 +1,61 @@
+﻿using System.Linq;
+using System.Web.Http;
+using System.Web.Http.Controllers;
+
+namespace D2L.Security.OAuth2.Authorization {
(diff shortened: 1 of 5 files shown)


src/D2L.Security.OAuth2.WebApi/Authorization/OAuth2AuthorizeAttribute.cs
```diff
@@ -0,0 +1,61 @@
...
+		private const string AUTH_HAS_RUN = "D2L.Security.OAuth2.Authorization.OAuth2AuthorizeAttribute_Auth_Has_Run";
+		private const string FAILING_ATTR_PROP = "D2L.Security.OAuth2.Authorization.OAuth2AuthorizeAttribute_Failing_Attribute";
+
+		protected override bool IsAuthorized( HttpActionContext actionContext ) {
+			var props = actionContext.Request.Properties;
+			if( props.ContainsKey( AUTH_HAS_RUN ) ) {
+				return true;
+			}
+
+			props.Add( AUTH_HAS_RUN, true );
```
j3parker: nit: nice if literal args have explicit arg names so I don't have to find docs ;) (lazy)

omsmith:
@j3parker okay as a temporary solution otherwise then?

PR #55, opened 2016-07-12
Update OAuth2AuthorizeAttribute.cs

lboyd:
Remove inheritance restriction to allow this attribute to be inherited from Abstract classes.
Resulted from investigation of : https://rally1.rallydev.com/#/23525809118d/detail/userstory/57136951529
[MetronAPI] Add new default scope for Metron API routes

### DIFF
--- a/src/D2L.Security.OAuth2.WebApi/Authorization/OAuth2AuthorizeAttribute.cs
+++ b/src/D2L.Security.OAuth2.WebApi/Authorization/OAuth2AuthorizeAttribute.cs
@@ -18,13 +18,13 @@ namespace D2L.Security.OAuth2.Authorization {
 
 			var actionAttributes = actionContext
 				.ActionDescriptor
-				.GetCustomAttributes<OAuth2AuthorizeAttribute>( inherit: false );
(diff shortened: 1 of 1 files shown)


lboyd:
Would be great to have @neverendingqs and @j3parker take a look if you have time...

j3parker:
Weird, I'm having trouble finding this on MSDN. The default is `true` I take it?

j3parker:
I can't find the source code for `System.Web.Http` though. I suppose if it works I'm cool with it. I'd usually just put `inherit: true` to avoid the confusion but nbd. Let me know if you've tried it and it works.

neverendingqs:
+1 on explicitly calling it out (don't see a downside with being more verbose here).

Found the abstract class: https://github.com/ASP-NET-MVC/aspnetwebstack/blob/master/src/System.Web.Http/Controllers/HttpActionDescriptor.cs#L161.

lboyd:
Yeah no problem from my side calling it out explicitly!

neverendingqs:
:shipit:

PR #56, opened 2016-07-12
Fix Permission Validation

mtjandra:
When validating multiple permissions, only the first permission was being validated. See added test with description `Extra permission is not granted` to understand what has been fixed with this change.

### DIFF
--- a/src/D2L.Security.OAuth2/Scopes/ScopeAuthorizer.cs
+++ b/src/D2L.Security.OAuth2/Scopes/ScopeAuthorizer.cs
@@ -30,12 +30,10 @@ namespace D2L.Security.OAuth2.Scopes {
 			var resourceScopes = groupScopes
 				.Where( s => IsMatch( s.Resource, requiredScope.Resource ) );
 
-			var permissionScopes = resourceScopes
-				.Where( s => s.Permissions.Any(
-					p => requiredScope.Permissions.All( rp => IsMatch( p, rp ) )
-				) );
+			var authorized = requiredScope.Permissions
+				.All( rp => resourceScopes.Any( r => r.Permissions.Any( p => IsMatch( p, rp )) ) );
 
-			return permissionScopes.Any();
+			return authorized;
 		}
 
 		private static bool IsMatch( string pattern, string actual ) {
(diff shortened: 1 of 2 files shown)


src/D2L.Security.OAuth2/Scopes/ScopeAuthorizer.cs
```diff
@@ -30,12 +30,10 @@ public static class ScopeAuthorizer {
 			var resourceScopes = groupScopes
 				.Where( s => IsMatch( s.Resource, requiredScope.Resource ) );
 
-			var permissionScopes = resourceScopes
-				.Where( s => s.Permissions.Any(
-					p => IsMatch( p, requiredScope.Permissions[0] )
-				) );
+			var authorized = requiredScope.Permissions
+				.All( rp => resourceScopes.Any( r => r.Permissions.Any( p => IsMatch( p, rp )) ) );
```
neverendingqs: Minor: Could use another variable to clarify it a bit:

```cs
var grantedPermissions = resourceScopes
    .Select( s => s.Permissions)
    .ToArray();

var authorized = requiredScope.Permissions
    .All( rp => grantedPermissions.Any( p => IsMatch( p, rp ) ) );
```

src/D2L.Security.OAuth2/Scopes/ScopeAuthorizer.cs
```diff
@@ -30,12 +30,13 @@ public static class ScopeAuthorizer {
 			var resourceScopes = groupScopes
 				.Where( s => IsMatch( s.Resource, requiredScope.Resource ) );
 
-			var permissionScopes = resourceScopes
-				.Where( s => s.Permissions.Any(
-					p => IsMatch( p, requiredScope.Permissions[0] )
-				) );
+			var grantedPermissions = resourceScopes
+				.SelectMany( s => s.Permissions );
```
neverendingqs: Consider using `ToArray()` avoids multiple enumerations.

omsmith:
On vacation, but I can look Monday. I thought it was understood this only did "singular" permissions, personally.

neverendingqs:
>On vacation, but I can look Monday. I thought it was understood this only did "singular" permissions, personally.
Related to if "g:r:p g2:r:p" is a "scope" or "scopes"?

mtjandra:
> On vacation, but I can look Monday. I thought it was understood this only did "singular" permissions, personally.

That is not how I understood it. The summary states '`Determines if granted scopes satisfy all required scopes`'. Also the scope class says '`Scope strings have the following representation: {group}:{resource}:{permission[,permission]*`'. So it seems that it should be able to validate the given 'scopes' (multiple permissions on some group:resource) based on the granted scopes.

omsmith:
LGTM

PR #57, opened 2016-08-08
US75330 - adding check for tokens with parts that are non-Base64Url encoded strings.

neverendingqs:
Workaround until https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/issues/482 is resolved.

### DIFF
--- a/src/D2L.Security.OAuth2/Validation/AccessTokens/AccessTokenValidator.cs
+++ b/src/D2L.Security.OAuth2/Validation/AccessTokens/AccessTokenValidator.cs
@@ -36,8 +36,14 @@ namespace D2L.Security.OAuth2.Validation.AccessTokens {
 			string token
 		) {
 			var tokenHandler = m_tokenHandler.Value;
+			var tokenInParts = token.Split( '.' );	// header.payload.signature
 
-			if( !tokenHandler.CanReadToken( token ) ) {
+			if(
+				!tokenHandler.CanReadToken( token ) ||
+				// https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/issues/482
+				!IsValidPart( tokenInParts[0] ) ||
+				!IsValidPart( tokenInParts[1] )
+			) {
 				throw new ValidationException( "Couldn't parse token" );
 			}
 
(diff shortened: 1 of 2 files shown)


src/D2L.Security.OAuth2/Validation/AccessTokens/AccessTokenValidator.cs
```diff
@@ -36,8 +36,14 @@ IPublicKeyProvider publicKeyProvider
...
+			var tokenInParts = token.Split( '.' );	// header.payload.signature
 
-			if( !tokenHandler.CanReadToken( token ) ) {
+			if(
+				!tokenHandler.CanReadToken( token ) ||
+				// https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/issues/482
+				!IsValidPart( tokenInParts[0] ) ||
+				!IsValidPart( tokenInParts[1] )
+			) {
 				throw new ValidationException( "Couldn't parse token" );
```
neverendingqs: Should this be a `InvalidTokenException`?

omsmith:
Is this something we actually ran into?

neverendingqs:
> Is this something we actually ran into?

Ran into it while testing changes for https://git.dev.d2l/projects/CORE/repos/lp/pull-requests/2978/diff#framework/web/D2L.LP.Web/Rest/Errors/ExceptionHandlers/OAuth2ValidationErrorHandler.cs

It was returning 500 on bad access tokens, and we wanted to change that to return 401. We did it by handling `ValidationException`s, but this issue causes `ArgumentException`s.

omsmith:
So, yeah I knew this was a deficiency of the `CanReadToken` (that it wouldn't catch invalid parts), but most libraries are the same way. Similarly, if the base64 isn't representing valid json it would also proceed to fail despite these checks saying it should be good.

This is now just re-doing some of the work (decoding and not using the output) that's going to be done over again in `ReadToken`. So, if anything I might suggest just not using `JwtSecurityTokenHandler` in this case if we're looking to handle these cases more specifically.

j3parker:
I'm not really worried about this error case because it's very obscure. This is a good example of test-driven-design leading to more complicated code with no pay-off. That being said, it degrades gracefully and there is a link in the code to the GitHub issue and its pretty obvious for how to test if this is still relevant so I'm not opposed to merging.

src/D2L.Security.OAuth2/Validation/AccessTokens/AccessTokenValidator.cs
j3parker: Seems right. You can fix that in a patch revision IMO (base class etc.) so that people can handle exceptions more specifically if they want.

neverendingqs:
The original reason for changing this was a UX concern (500 on certain malformed access tokens). After another discussion with the PO, we're okay if this case (non-Base64Url encoded strings) isn't covered.

Since we're adding complexity with little gain, I'm opting to just close this PR without merging unless there are any concerns.

j3parker:
Yeah, it's hard to imagine why this would happen in practice. If `JwtSecurityTokenHandler`s behaviour changes we'll get it covered for free anyway.

brentschmaltz:
@neverendingqs @j3parker nice catch. We will touch this up in our next minor release.

PR #61, opened 2016-11-18
Added MonoRsaSecurityKey to allow use of library within mono

drose-d2l:


src/D2L.Security.OAuth2/Keys/Default/RsaJsonWebKey.cs
```diff
@@ -79,6 +79,9 @@ string e
 				keyFactory: () => {
 					var rsa = new RSACryptoServiceProvider() { PersistKeyInCsp = false };
 					rsa.ImportParameters( m_parameters );
+#if __MonoCS__ || MONO
+					var key = new MonoRsaSecurityKey( rsa );
+#endif
```
omsmith: Should there not be an `#elif` on these?

omsmith:
Do you know _why_ Mono is different?

PR #62, opened 2017-02-21
US81987: exposing some AccessTokenExtensions methods for use in platf…

astadelbauer:
…ormTools/impersonation


j3parker:
I'll bump the version and open a PR to use the new version in the LMS after CI passes & this is merged.

omsmith:
Would we not just use `ID2LPrincipal`?

astadelbauer:
ID2LPrincipal would be great, but we didn't think it would work since we're passing the token in the form post body and not actually doing authentication with it -- @j3parker understands more than I do :)

omsmith:
Oh yeah, that part.

PR #63, opened 2017-03-14
Littlefoot/statics/bump nunit and http mock version

LiyueWang:


src/D2L.Security.OAuth2.TestFramework/Properties/AssemblyInfo.cs
```diff
@@ -32,8 +32,8 @@
 // You can specify all the values or you can default the Build and Revision Numbers
 // by using the '*' as shown below:
 // [assembly: AssemblyVersion("1.0.*")]
-[assembly: AssemblyVersion("5.2.1.0")]
-[assembly: AssemblyFileVersion("5.2.1.0")]
+[assembly: AssemblyVersion("5.2.2.0")]
```
omsmith: I would recommend against updating version numbers within the PR

test/D2L.Security.OAuth2.Benchmarks/App.config
```diff
@@ -9,6 +9,14 @@
         <assemblyIdentity name="nunit.framework" publicKeyToken="96d09a1eb7f44a77" culture="neutral" />
         <bindingRedirect oldVersion="0.0.0.0-2.6.4.14350" newVersion="2.6.4.14350" />
       </dependentAssembly>
+      <dependentAssembly>
+        <assemblyIdentity name="log4net" publicKeyToken="669e0ddf0bb1aa2a" culture="neutral" />
+        <bindingRedirect oldVersion="0.0.0.0-1.2.15.0" newVersion="1.2.15.0" />
+      </dependentAssembly>
+      <dependentAssembly>
+        <assemblyIdentity name="nunit.framework" publicKeyToken="2638cd05610744eb" culture="neutral" />
+        <bindingRedirect oldVersion="0.0.0.0-3.6.1.0" newVersion="3.6.1.0" />
```
omsmith: Seems a little in conflict with the previous _nunit.framework_ entry

src/D2L.Security.OAuth2.TestFramework/Properties/AssemblyInfo.cs
LiyueWang: gotcha, thanks!

test/D2L.Security.OAuth2.Benchmarks/App.config
LiyueWang: yes. and it seems if I want to upgrade the NUnit(from 2.6.4 to 3.6.1) version for D2L.Security.OAuth2.TestFramework, then I also need to upgrade for the tests projects D2L.Security.OAuth2.Benchmarks, D2L.Security.OAuth2.IntegrationTests, D2L.Security.OAuth2.UnitTests, right?

And in order to upgrade NUnit version from 2.6.4 to 3.6.1, I also have to upgrade HttpMock from 1.1.5 to 1.3.3, log4net from 2.0.3 to 2.0.8.  as the dependencies NUnit->HttpMock->log4net.

I will upgrade all of them. 
The reason I want to do this is because I am upgrading Nunit from 2.6.4 to 3.6.1 for all test projects in Core. And one of the project depends on D2L.Security.OAuth2.TestFramework with old version of Nunit and HttpMock, which will cause the build failed.

I am not quite familiar with the usage of D2L.Security.OAuth2, if this will cause any other problems, please let me know. Thanks!

appveyor.yml
```diff
@@ -14,3 +14,9 @@ build:
 configuration:
 - Debug
 - Release
+test:
+  assemblies:
```
omsmith: Any way to avoid this? Seems like if we added a test assembly and forgot to add it here, we would be none the wiser?

LiyueWang: This is to ignore HttpMock.dll. Since Nunit 3 has problem to process with HttpMock.dll, I am not sure why,  try to figure out

LiyueWang: I didn't figure out why HttpMock 1.3.3 will get `Has no TestFixtures`, but the old version are fine. And it seems there are a few  `Has no TestFixtures` issues in nunit3. Just open a new issue about this one. 

But still, is HttpMock.dll supposed to be a test assembly to run the tests, in fact there is no tests in it. 
>Tests run: 0, Errors: 0, Failures: 0, Inconclusive: 0, Time: 0.00674923043719606 seconds
  Not run: 0, Invalid: 0, Ignored: 0, Skipped: 0
And I assume we don't care if we run HttpMock.dll, right?

And unfortunately, it is impossible  to exclude a specific assembly in AppVeyor...

LiyueWang: Response from Nunit authors: https://github.com/nunit/nunit/issues/2082
This is a different behavior between nunit 2 runner and nunit 3 runner.
With nunit 3 runner, if an assembly DOES reference nunit.framework (as HttpMock.dll does), the assembly should HAVE a NonTestAssemblyAttribute (HttpMock doesn't have this attribute), and in the command-line with `--skipnontestassemblies`, then the runner will pass it. Otherwise, it is treated as INVALID.

So for now the only way I can do is to use the appveyor config file to skip the HttpMock.dll.

LiyueWang: @omsmith if no objection, I will merge this PR.

omsmith: Seems like what we have to do at the moment

PR #65, opened 2017-04-07
Add TravisCI badge for Mono build

j3parker:
Not sure if we're going to go this route.

If we do we'd need another job that could combine the two build artifacts into one glorious nupkg and publish it.


omsmith:
Not much to review here :P

j3parker:
Yeah I guess I was using that to notify you but that's annoying because it shows up in your list

PR #69, opened 2017-08-03
Add support for bare JWK to IJwkProvider

j3parker:
This adds a new method IJwkProvider.RequestJwkAsync. It falls back to
RequestJwksAsync on 404. We want to move to JWK to support better
caching (i.e. performance) and it is more compatible with services
like S3+CF that would increase our fault-tolerance.

### DIFF
--- a/src/D2L.Security.OAuth2/Keys/Default/Data/JwksProvider.cs
+++ b/src/D2L.Security.OAuth2/Keys/Default/Data/JwksProvider.cs
@@ -1,4 +1,5 @@
 ﻿using System;
+using System.Net;
 using System.Net.Http;
 using System.Threading.Tasks;
 using D2L.Security.OAuth2.Validation.Exceptions;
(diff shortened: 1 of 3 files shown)


j3parker:
No tests yet, just wanted to get your thoughts.

I'm thinking we could plumb an optional boolean through the validation stuff to enable this code path. Step 1 would be to enable it in the LMS because it would be totally safe (auth supports JWK.)

src/D2L.Security.OAuth2/Keys/Default/Data/JwksProvider.cs
```diff
@@ -1,65 +1,118 @@
...
+			if( !allKeys.TryGetKey( keyId, out theKey ) ) {
+				throw new PublicKeyLookupFailureException(
+					$"couldn't find public key {keyId} after falling back to jwks route"
+				);
+			}
+
+			return theKey;
+		}
+
+		string IJwksProvider.Namespace => m_authEndpoint.AbsoluteUri;
```
omsmith: Is this the same as `{ get; } =`, or can you not do that because not constant?

src/D2L.Security.OAuth2/Keys/Default/Data/JwksProvider.cs
```diff
@@ -1,65 +1,118 @@
...
+
+		private static Uri GetJwksEndpoint( Uri authEndpoint ) {
+			string authRoot = MakeSureThereIsATrailingSpace( authEndpoint );
+
+			authRoot += ".well-known/jwks";
+
+			return new Uri( authRoot );
+		}
+
+		private static string MakeSureThereIsATrailingSpace( Uri uri ) {
```
omsmith: TrailingSlash?

src/D2L.Security.OAuth2/Keys/Default/Data/JwksProvider.cs
j3parker: `{ get; } =` would effectively hold a copy of that string (which is maybe just a reference internally?) i.e. it's an auto-property and equivalent to `{ get; }` with assignment in the constructor. This is equivalent to what preceded it - a regular property (no extra state.)

I don't really have a preference, I just unnecessarily applied the fixit :P

omsmith:
I'll continue reviewing this, once you review https://github.com/Brightspace/D2L.Security.OAuth2/pull/67 ;)

j3parker:
Haha good call

j3parker:
Bumpity

src/D2L.Security.OAuth2/Keys/Default/Data/JwksProvider.cs
```diff
@@ -1,65 +1,118 @@
...
+			string authRoot = MakeSureThereIsATrailingSlash( authEndpoint );
+
+			authRoot += ".well-known/jwks";
+
+			return new Uri( authRoot );
+		}
+
+		private static string MakeSureThereIsATrailingSlash( Uri uri ) {
+			string root = uri.ToString();
+			if ( root[root.Length - 1 ] == '/') {
```
omsmith: Is it possible for `root.Length == 0`? Probably not, given we had a `Uri`

j3parker: Looking through the source code it could happen in irrelevant cases but not when there is a host (the only case we'll care about.)
I think I'm ok with an NRE in that case... good catch though.

PR #73, opened 2017-11-13
Making Scope immutable

JeffAshton:


test/D2L.Security.OAuth2.UnitTests/Scopes/ScopeTests.cs
```diff
@@ -40,7 +40,7 @@ internal sealed class ScopeTests {
 			isParsed.Should().BeTrue();
 			scope.Group.Should().Be( group );
 			scope.Resource.Should().Be( resource );
-			scope.Permissions.ShouldAllBeEquivalentTo( permissions );
+			CollectionAssert.AreEquivalent( permissions, scope.Permissions );
```
JeffAshton: The fluent library has a bug, so just using the native NUnit assertions.

```
System.InvalidOperationException : No generic method 'ToArray' on type 'System.Linq.Enumerable' is compatible with the supplied type arguments and arguments. No type arguments should be provided if the method is non-generic. 
   at System.Linq.Expressions.Expression.FindMethod(Type type, String methodName, Type[] typeArgs, Expression[] args, BindingFlags flags)
```

https://ci.appveyor.com/project/Brightspace/d2l-security-oauth2/build/5.2.0.88-ltruxvag/job/c9d7ig76jrxy13o8#L156

PR #77, opened 2018-03-01
Update appveyor.yml

toddlang:


j3parker:
wtf @ the CI output?

omsmith:
Tried re-running it, doesn't seem to have helped. If I had to guess, we're referencing some sort of package that isn't available on nuget.org, but that doesn't seem to be that case?

j3parker:
It's weird, the build is passing without this change. What would be different on VS2017?

toddlang:
Log looks like some kind of inability to get the required permissions to pull packages, no?

PR #79, opened 2018-04-02
Improve error message when token is not yet valid

j3parker:
I didn't make a new exception type. I looked around and the utility appears to be for tests. If you'd like a test I'll redo this with a custom exception type.


src/D2L.Security.OAuth2/Validation/AccessTokens/AccessTokenValidator.cs
```diff
@@ -84,10 +84,12 @@ string token
 					token,
 					validationParameters,
 					out securityToken
-        );
+        		);
 				accessToken = new AccessToken( ( JwtSecurityToken )securityToken );
 			} catch( SecurityTokenExpiredException e ) {
 				throw new ExpiredTokenException( e );
+			} catch( SecurityTokenNotYetValidException e ) {
+				throw new ValidationException( "Token was from the future (nbf)", e );
```
omsmith: nit: expired token exception message is in present tense. _The access token is expired_.

j3parker: Good call that read pretty weird

PR #80, opened 2018-05-31
Update to IdentityModel 5.2.2

omsmith:
There's other bits and bobs that could be updated (using their
JsonWebKey implementations, for instance), but I feel best / safest
about this step.

I feel real bad. I did a 'Format All Files' to clean up a Find and Replace I
did on the old namespace usings, and... yeah :(

https://github.com/Brightspace/D2L.Security.OAuth2/pull/80/files?w=1


omsmith:
# Before
before

# After
after

PR #82, opened 2018-06-05
Throw on scope parse

omsmith:


omsmith:
Hmm, build failure is it not understanding the newer syntax

src/D2L.Security.OAuth2/Scopes/Scope.cs
```diff
@@ -124,39 +124,38 @@ public Scope( string group, string resource, string permission )
 		/// *:*:read                    => Read only for all resources in all groups
 		/// </example>
 		public static bool TryParse( string scopePattern, out Scope scope ) {
+			try {
+				scope = Parse( scopePattern );
+				return true;
+			} catch {
```
JeffAshton: :(  Not a blocker, but I don't love it

src/D2L.Security.OAuth2/Scopes/Scope.cs
```diff
@@ -124,39 +124,38 @@ public Scope( string group, string resource, string permission )
...
-			scope = null;
-
+		/// <summary>
+		/// Parses a scope pattern into a <see cref="Scope"/>.
+		/// </summary>
+		/// <param name="scopePattern">The scope pattern in the form of
+		/// group:resource:permission[,permission]*.</param>
+		/// <returns>If the scope is formatted correctly, returns the <see cref="Scope"/> object</returns>
+		/// <exception cref="ArgumentException"></exception>
+		public static Scope Parse( string scopePattern ) {
```
JeffAshton: Consider renaming to ```ParseOrThrow``` to make it easier to find the breaking changes

JeffAshton: Given the dirty ```null``` check that already existed above

src/D2L.Security.OAuth2/Scopes/Scope.cs
omsmith: What would you rather? Just catch ArgumentException?

src/D2L.Security.OAuth2/Scopes/Scope.cs
omsmith: I'm not super-worried about that. There could be other-other third party users of this library of course, but otherwise the only usages for us that use `Scope.Parse` on user-supplied data were withing the `GetScopes` and `GetGrantedScopes` extensions that I updated in this PR. All other usages are on constants in code.

JeffAshton: Cool, if you know that to be true, then I'm content

src/D2L.Security.OAuth2/Scopes/Scope.cs
JeffAshton: No, I just hate exceptions.  It's fine if this is all mostly constants

PR #91, opened 2018-08-14
Marking ITokenSigner as Immutable

toddlang:


src/D2L.Security.OAuth2/Keys/Default/RotatingPrivateKeyProvider.cs
```diff
@@ -13,8 +12,16 @@ internal sealed class RotatingPrivateKeyProvider : IPrivateKeyProvider {
 		private readonly IDateTimeProvider m_dateTimeProvider;
 		private readonly TimeSpan m_keyRotationPeriod;
 
+		[Mutability.Audited(
+			"Todd Lang",
+			"14-Aug-2018",
+			".Net class we can't mark, used internally without mutability." )]
```
daryl-mcmillan: I'm guessing we do mutate this class, otherwise it's not very useful

src/D2L.Security.OAuth2/Keys/Default/EcDsaPrivateKeyProvider.cs
```diff
@@ -2,11 +2,16 @@
...
+using D2L.CodeStyle.Annotations;
 
 namespace D2L.Security.OAuth2.Keys.Default {
 	internal sealed partial class EcDsaPrivateKeyProvider : IPrivateKeyProvider {
 
 		private readonly ID2LSecurityTokenFactory m_d2lSecurityTokenFactory;
+		[Mutability.Audited(
+			"Todd Lang",
+			"14-Aug-2018",
+			".Net class we can't mark.")]
```
ChrisAtD2L: If this is a .Net class that is immutable, why did it not join all the other .Net classes in the exemptions list in the analyzer code?
https://github.com/Brightspace/D2L.CodeStyle/blob/master/src/D2L.CodeStyle.Analyzers/Immutability/KnownImmutableTypes.cs

src/D2L.Security.OAuth2/Keys/Default/RotatingPrivateKeyProvider.cs
ChrisAtD2L: Unlike `CngAlgorithm`, I think this one makes sense to audit.  We don't want to automatically assume that every SemaphoreSlim is safe.  We should probably still manually inspect these to make sure that they are still sane in a multi-tenant environment.

src/D2L.Security.OAuth2/Keys/Development/StaticPrivateKeyProvider.cs
```diff
@@ -12,6 +13,10 @@ namespace D2L.Security.OAuth2.Keys.Development {
 	[Obsolete("Only use this in tests and for prototyping.")]
 	internal sealed class StaticPrivateKeyProvider : IPrivateKeyProvider {
 		private readonly Guid m_keyId;
+		[Mutability.Audited(
+			"Todd Lang",
+			"14-Aug-2018",
+			".Net class that is not mutable but can't be marked.")]
```
ChrisAtD2L: Candidate for KnownImmutableTypes in the analyzer:
https://github.com/Brightspace/D2L.CodeStyle/blob/master/src/D2L.CodeStyle.Analyzers/Immutability/KnownImmutableTypes.cs

src/D2L.Security.OAuth2/Keys/Development/InMemoryPublicKeyDataProvider.cs
```diff
@@ -12,6 +13,10 @@ namespace D2L.Security.OAuth2.Keys.Development {
 	/// </summary>
 	[Obsolete("Only use this in tests and for prototyping without a db")]
 	public sealed class InMemoryPublicKeyDataProvider : IPublicKeyDataProvider {
+		[Mutability.Audited(
+			"Todd Lang",
+			"14-Aug-2018",
+			"Only used in tests, mutability not a concern here.")]
```
ChrisAtD2L: If this is only used in tests, why not move it to a project that uses the less strict TestAnalyzer?  I feel like we already ignore test-only things in other assemblies without resorting to `Mutability.Audited`

PR #96, opened 2018-08-29
Rolling back to net461

toddlang:


neverendingqs:
Curious: what was wrong with 4.7

omsmith:
I assume it has something to do with us trying to update the package in the auth service, and not forcing us to upgrade the .NET framework there

j3parker:
An important service that uses this library isn't on that version of the framework yet and we need to make some changes. Nothing required this being updated to the new framework at this time.

PR #100, opened 2018-09-04
Omsmith/jwks fallback all errors

omsmith:
BDP's S3 bucket doesn't yet provide `/jwk/:id` and 403s on unknown s3 keys.

### DIFF
--- a/test/D2L.Security.OAuth2.IntegrationTests/Keys/Default/Data/JwksProviderTests.cs
+++ b/test/D2L.Security.OAuth2.IntegrationTests/Keys/Default/Data/JwksProviderTests.cs
@@ -164,5 +164,27 @@ namespace D2L.Security.OAuth2.Keys.Default.Data {
 				Assert.AreEqual( GOOD_JWK_ID, jwk.Id );
 			}
 		}
+
+		[Test]
+		public async Task RequestJwkAsync_500_Fallback_Success() {
+			using( SetupJwkServer( out string host, hasJwk: false, jwkStatusCode: HttpStatusCode.InternalServerError ) )
+			using( HttpClient httpClient = new HttpClient() ) {
+				IJwksProvider jwksProvider = new JwksProvider(
+					httpClient,
+					new Uri( host + GOOD_PATH )
+				);
+
+				JsonWebKeySet jwks = await jwksProvider
+					.RequestJwkAsync( GOOD_JWK_ID )
+					.SafeAsync();
+				Assert.IsNotNull( jwks );
+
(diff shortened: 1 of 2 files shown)


j3parker:
300s? Maybe we follow redirects though.

test/D2L.Security.OAuth2.IntegrationTests/Keys/Default/Data/JwksProviderTests.cs
```diff
@@ -20,91 +20,171 @@ public class JwksProviderTests {
...
+			}
 		}
 
 		[Test]
-		public void RequestJwksAsync_CantReachServer_Throws() {
-			IJwksProvider publicKeyProvider = new JwksProvider(
-				new HttpClient(),
-				new Uri( "http://foo.bar.fakesite.isurehopethisisneveravalidTLD" )
-			);
+		public async Task RequestJwkAsync_Success() {
```
travisMiehm: nit: the next three tests could be `TestCase`s

j3parker:
@neverendingqs will this be supported in the near future?

test/D2L.Security.OAuth2.IntegrationTests/Keys/Default/Data/JwksProviderTests.cs
```diff
@@ -20,91 +20,171 @@ public class JwksProviderTests {
...
-				JsonWebKeySet jwks = await publicKeyProvider
-				.RequestJwksAsync()
-				.SafeAsync();
-			} );
+				JsonWebKeySet jwks = await jwksProvider
+					.RequestJwkAsync( GOOD_JWK_ID )
+					.SafeAsync();
+				Assert.IsNotNull( jwks );
+
+				//jwksServer.AssertWasCalled( x => x.Get( GOOD_JWK_PATH ) );
```
travisMiehm: nit: remove commented code

neverendingqs:
Related: https://github.com/Brightspace/bdp-auth-infrastructure/issues/22. It's just making another prefix public, but not yet implemented.

omsmith:
Yeah, don't believe the 300 gets exposed to us

test/D2L.Security.OAuth2.IntegrationTests/Keys/Default/Data/JwksProviderTests.cs
omsmith: Yeah, I wanted these calls to work, but my local was having some error loading nunit and was hoping someone smarter would fix it for me- perhaps the library uses a different version for its assertions and we don't have a binding redirect or whatever.

PR #107, opened 2018-12-05
refact: DateTime => DateTimeOffset

omsmith:


src/D2L.Security.OAuth2/Utilities/DateTimeHelpers.cs
```diff
@@ -1,11 +0,0 @@
-﻿using System;
-
-namespace D2L.Security.OAuth2 {
-	internal static class DateTimeHelpers {
-		public static readonly DateTime EPOCH = new DateTime( 1970, 1, 1, 0, 0, 0, 0, DateTimeKind.Utc );
-
-		public static DateTime FromUnixTime( long seconds ) {
```
j3parker: Thanks for cleaning this up too

j3parker:
Good stuff. There is so much code where we got this wrong.

PR #110, opened 2019-01-04
Remove fallback to jwks when fetching jwk

j3parker:
* Everyone supports JWK
* The fallback can cause perf issues due to the response size of jwks, its uncachability and a bug with one of the public key hosts in the system
* Let's keep everyone supporting jwk

### DIFF
--- a/src/D2L.Security.OAuth2.TestFramework/AuthServiceMock.cs
+++ b/src/D2L.Security.OAuth2.TestFramework/AuthServiceMock.cs
@@ -94,7 +94,7 @@ namespace D2L.Security.OAuth2.TestFramework {
 				keyDtos.Add( dto );
 
 				m_server
-					.Stub( r => r.Get( $"jwk/{ key.Id }" ) )
+					.Stub( r => r.Get( $"/jwk/{ key.Id }" ) )
 					.Return( JsonConvert.SerializeObject( dto ) )
 					.OK();
 			}
--- a/src/D2L.Security.OAuth2/Keys/Default/Data/JwksProvider.cs
+++ b/src/D2L.Security.OAuth2/Keys/Default/Data/JwksProvider.cs
@@ -39,18 +39,8 @@ namespace D2L.Security.OAuth2.Keys.Default.Data {
 			var url = GetJwkEndpoint( m_authEndpoint, keyId );
 			try {
 				using( var res = await m_httpClient.GetAsync( url ).SafeAsync() ) {
-					// This is temporary while we try to fully deprecate the
-					// JWKS route. 404 might mean the key doesn't exist (which
-					// will make the call to jwks likely return 200 but still
-					// result in no key - that's fine but slow) or the jwk
-					// route isn't supported which will make this recover (but
-					// slowely.) Where it matters (the LMS and auth) JWK is
-					// supported so this shouldn't be necessary. We don't
-					// expect many "legitimate" 404s (keys that don't exist.)
-					// so in practice this should only happen when it's
-					// actually important, if it ever happens.
-					if( res.StatusCode != HttpStatusCode.OK ) {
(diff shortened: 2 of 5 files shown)


src/D2L.Security.OAuth2.TestFramework/AuthServiceMock.cs
```diff
@@ -94,7 +94,7 @@ public enum KeyType {
 				keyDtos.Add( dto );
 
 				m_server
-					.Stub( r => r.Get( $"jwk/{ key.Id }" ) )
+					.Stub( r => r.Get( $"/jwk/{ key.Id }" ) )
```
j3parker: This was broken - but tests were falling back to jwks :)

src/D2L.Security.OAuth2/Keys/Default/Data/JwksProvider.cs
```diff
@@ -39,18 +39,8 @@ Uri authEndpoint
...
-					// route isn't supported which will make this recover (but
-					// slowely.) Where it matters (the LMS and auth) JWK is
-					// supported so this shouldn't be necessary. We don't
-					// expect many "legitimate" 404s (keys that don't exist.)
-					// so in practice this should only happen when it's
-					// actually important, if it ever happens.
-					if( res.StatusCode != HttpStatusCode.OK ) {
-						return await ( this as IJwksProvider ).RequestJwksAsync().SafeAsync();
+					if ( res.StatusCode == HttpStatusCode.NotFound ) {
+						return JsonWebKeySet.Empty( url );
```
j3parker: This interface returns a `JsonWebKeySet`, usually a single key. To keep the exception messages correct it was sanest to return an empty JWKS here, and let the thing that notices that and throws `PublicKeyNotFoundException` do its thing.

src/D2L.Security.OAuth2/Keys/Default/JsonWebKeySet.cs
```diff
@@ -34,9 +34,19 @@ internal sealed class JsonWebKeySet {
...
-			m_keys = ImmutableArray.Create( jsonWebKey );
+		internal JsonWebKeySet( JsonWebKey jsonWebKey, Uri src )
+			: this(
+				ImmutableArray.Create( jsonWebKey ),
+				src
+			) { }
+
+		private JsonWebKeySet( ImmutableArray<JsonWebKey> keys, Uri src ) {
+			m_keys = keys;
+			Source = src ?? throw new NotImplementedException( nameof( src ) );
```
travisMiehm: `ArgumentNullException`?

j3parker: Oh shit thanks, auto-complete fail

j3parker:
> Verified okay for existed auth service clients?

Yep, a bit ago, but I'll revalidate today.

> Also, we'll need to make a branch and cherry-pick this over, master isn't shippable (it is theoretically, but not practically)

D'oh, ok. Do you know what commit we should branch at?

omsmith:
Likely candidates:
* 5f7e2b410bcb5794e3a5942b5edac78e534c6ea8 (7.2.2)
* a85271cffd78d4cb4036fb34949c587e430947fd (7.2.2 formatting / code cleanups)

j3parker:
I'll go with a85271c to avoid possible merge conflicts

omsmith:
Made a branch [7.x](https://github.com/Brightspace/D2L.Security.OAuth2/commits/7.x) with https://github.com/Brightspace/D2L.Security.OAuth2/commit/58d78092d914ac5f76affde9b364825054f263b8 cherry-picked. These apply cleanly on top.

j3parker:
I can cherry-pick this tomorrow and we'll figure out how exactly we want to do this.

PR #113, opened 2019-03-21
Allow TokenSigner to sign tokens with complex claims

khorwood:
Adds support for signing JWT messages with complex claims such as
objects or arrays for use by integrations such as LTI v1.3 launch
workflow.


khorwood:
Further updates expected in additional PRs:
- Support arbitrary string scopes in addition to D2L scopes (eg. LTI URI based scopes)
- Support arbitrary string `kid` from partner supplied jwks URIs

omsmith:
I believe the only user of `ITokenSigner` is https://github.com/Brightspace/D2L.Security.OAuth2/blob/7.x/src/D2L.Security.OAuth2/Provisioning/Default/AccessTokenProvider.cs. So `UnsignedToken` _could_ just be updated, instead of providing two variations.

omsmith:
> Support arbitrary string kid from partner supplied jwks URIs

Because OIDC?

omsmith:
@j3parker If we're doing to support inter-operating with other folks JWKS, then dc85894ba1c8e60076c85c7c4358037888b2e4c8 becomes invalid.

omsmith:
(And fetching `/jwk/:id` in the first place becomes confusing)

omsmith:
@khorwood We have some d2l-isms in this library. I wonder if

> Support arbitrary string kid from partner supplied jwks URIs

Is better left to using a "standard" OIDC implementation to validate incoming tokens, and then hand it off into our stack?

khorwood:
> 
> 
> I believe the only user of `ITokenSigner` is [AccessTokenProvider](https://github.com/Brightspace/D2L.Security.OAuth2/blob/7.x/src/D2L.Security.OAuth2/Provisioning/Default/AccessTokenProvider.cs). So `UnsignedToken` _could_ just be updated, instead of providing two variations.

Could do, would need to add an extra Provision method to support the funky LTI claims needed which aren't compatible with the `Claims` class. Otherwise we could leverage the `ITokenSigner` directly from the LMS launch flow.

khorwood:
> 
> 
> @khorwood We have some d2l-isms in this library. I wonder if
> 
> > Support arbitrary string kid from partner supplied jwks URIs
> 
> Is better left to using a "standard" OIDC implementation to validate incoming tokens, and then hand it off into our stack?

I'd like to, but we'll also be supporting a Service-to-Service flow where the Tool (registered as an app in AuthService) will be sending signed assertions which we'll need to verify before issuing an Auth Service token to them.

omsmith:
> I'd like to, but we'll also be supporting a Service-to-Service flow where the Tool (registered as an app in AuthService) will be sending signed assertions which we'll need to verify before issuing an Auth Service token to them.

Right, which is where I was suggesting the standard OIDC implementation would validate that incoming assertion, and after validating, then giving one of our tokens. i.e. send them to a different endpoint?

khorwood:
> 
> 
> > I'd like to, but we'll also be supporting a Service-to-Service flow where the Tool (registered as an app in AuthService) will be sending signed assertions which we'll need to verify before issuing an Auth Service token to them.
> 
> Right, which is where I was suggesting the standard OIDC implementation would validate that incoming assertion, and after validating, then giving one of our tokens. i.e. send them to a different endpoint?

The LTI launch is inspired by OIDC, and doesn't quite follow the spec, so we'll be implementing specific routes in the LMS to handle the launch communication. This specific PR is just about giving us the ability to sign the message using the LMS's PK.

src/D2L.Security.OAuth2/Keys/UnsignedComplexToken.cs
```diff
@@ -0,0 +1,72 @@
...
+    /// A token that has yet to be signed, supports complex claims
+    /// </summary>
+    public sealed class UnsignedComplexToken {
+        private readonly string m_issuer;
+        private readonly string m_audience;
+        private readonly IReadOnlyDictionary<string, object> m_claims;
+        private readonly DateTime m_notBefore;
+        private readonly DateTime m_expiresAt;
+
+        /// <summary>
```
j3parker: nit: whitespace

j3parker:
Sorry for not getting to this sooner 😞 

What did you think about @omsmith 's idea about modifying `UnsignedToken`?

These changes are required for LTI message right? So like, only in the LMS? I wonder it might be easier/faster to avoid using D2L.Security.OAuth2 to sign/handle the tokens, and just provide our credentials to the underlying libraries. Just a thought - I don't want to distract you too much.

khorwood:
> 
> 
> Sorry for not getting to this sooner 😞
> 
> What did you think about @omsmith 's idea about modifying `UnsignedToken` vs a second type?
> 
> These changes are required for LTI message right? So like, only in the LMS? I wonder it might be easier/faster to avoid using D2L.Security.OAuth2 to sign/handle the tokens, and just provide our credentials to the underlying libraries. Just a thought - I don't want to distract you too much.

Yeah, we only need the funky claims for the LTI message to the tool.

I'm open to condensing the class, I'm just not sure of the optimal approach. Which would be better in your eyes?
- Add an additional `IReadOnlyDictionary<string, object>` member to contain the complex claims, different constructors, and add logic to use one *OR* the other
- Replace the `ReadOnlyCollection<Claim>` with `IReadOnlyDictionary<string, object>`, adding a new constructor and preserving the old for backward compatibility

omsmith:
Shouldn't need to worry about any backwards compatibility if you also update AccessTokenProvider

khorwood:
Revised as per your suggestions. I noticed a lot of whitespace changes. Is there an `.editorconfig` I should be using for this repo?

omsmith:
Not presently. I can add one though

omsmith:
@khorwood I added an .editorconfig. I don't know that it will perfectly represent the current state, but I did my best.

src/D2L.Security.OAuth2/Keys/Default/TokenSigner.cs
```diff
@@ -1,39 +1,49 @@
...
+        async Task<string> ITokenSigner.SignAsync( UnsignedToken token ) {
+            JwtSecurityToken jwt;
+            using( D2LSecurityToken securityToken = await m_privateKeyProvider
+                .GetSigningCredentialsAsync()
+                .SafeAsync()
+            ) {
+                jwt = new JwtSecurityToken(
+                    issuer: token.Issuer,
+                    audience: token.Audience,
+                    claims: new List<Claim>(),
```
omsmith: `Enumerable.Empty<Claim>()`

src/D2L.Security.OAuth2/Keys/Default/TokenSigner.cs
```diff
@@ -1,39 +1,49 @@
...
+                    notBefore: token.NotBefore,
+                    expires: token.ExpiresAt,
+                    signingCredentials: securityToken.GetSigningCredentials()
+                );
+
+                var claims = token.Claims;
+                foreach( var claim in claims ) {
+                    if( jwt.Payload.ContainsKey( claim.Key ) ) {
```
omsmith: This can't really occur _within_ the dictionary, so this just becomes "built-in" claims like `aud`, `iss`, `exp` etc. I think this should probably throw. What would happen before?

src/D2L.Security.OAuth2/Keys/UnsignedToken.cs
```diff
@@ -1,75 +1,97 @@
...
+        public UnsignedToken(
+            string issuer,
+            string audience,
+            IList<Claim> claims,
+            DateTime notBefore,
+            DateTime expiresAt
+        ) {
+            m_issuer = issuer;
+            m_audience = audience;
+            m_claims = claims.ToDictionary( t => t.Type, t => (object)t.Value );
```
omsmith: I think I imagined this constructor going away. This is maybe fine. Just noting this could throw a "key already added" exception - avoiding having to do this conversion ourselves removes the chance of that "oops" case from our API surface. No particularly strong feelings though.

src/D2L.Security.OAuth2/Keys/Default/TokenSigner.cs
khorwood: It was throwing at this point in some of the unit tests, hence the dupe check.

src/D2L.Security.OAuth2/Keys/UnsignedToken.cs
khorwood: Will drop the old constructor.

src/D2L.Security.OAuth2/Keys/UnsignedToken.cs
```diff
@@ -11,7 +9,7 @@ namespace D2L.Security.OAuth2.Keys {
 	public sealed class UnsignedToken {
 		private readonly string m_issuer;
 		private readonly string m_audience;
-		private readonly ReadOnlyCollection<Claim> m_claims;
```
j3parker: So I take it we were misusing the `Claim` type (from System.Security.Claims, not the JWT stuff) here?

khorwood: Unsure, the `Claim` type was valid for use with the `JwtSecurityToken`, but Microsoft's implementation was strict around not allowing claims to be nested or contain arrays.

khorwood: At least not through manual construction.. It supported a BinaryReader which could have in theory allowed for anything.

PR #114, opened 2019-05-09
Add a standard JWKS provider for 3rd party support

khorwood:
Adds a standard JWKS provider, and changes the keyId data type from Guid
to string to allow for arbitrary kid values.
Adds a new factory method to AccessTokenValidatorFactory to support
validation using the standard JWKS provider.

Updated the changes in https://git.dev.d2l/projects/AUTH/repos/auth-service/pull-requests/490 to pivot to this approach.

### DIFF
--- a/src/D2L.Security.OAuth2/Keys/Default/Data/D2LJwksProvider.cs
+++ b/src/D2L.Security.OAuth2/Keys/Default/Data/D2LJwksProvider.cs
@@ -2,15 +2,16 @@
 using System.Net;
 using System.Net.Http;
 using System.Threading.Tasks;
+using System.Web;
 using D2L.Security.OAuth2.Validation.Exceptions;
 using D2L.Services;
 
 namespace D2L.Security.OAuth2.Keys.Default.Data {
-	internal sealed class JwksProvider : IJwksProvider {
+	internal sealed class D2LJwksProvider : IJwksProvider {
 		private readonly HttpClient m_httpClient;
 		private readonly Uri m_authEndpoint;
 
-		public JwksProvider(
+		public D2LJwksProvider(
 			HttpClient httpClient,
 			Uri authEndpoint
 		) {
@@ -35,7 +36,7 @@ namespace D2L.Security.OAuth2.Keys.Default.Data {
 			}
 		}
 
-		async Task<JsonWebKeySet> IJwksProvider.RequestJwkAsync( Guid keyId ) {
+		async Task<JsonWebKeySet> IJwksProvider.RequestJwkAsync( string keyId ) {
 			var url = GetJwkEndpoint( m_authEndpoint, keyId );
 			try {
 				using( var res = await m_httpClient.GetAsync( url ).SafeAsync() ) {
@@ -77,10 +78,10 @@ namespace D2L.Security.OAuth2.Keys.Default.Data {
 			return new PublicKeyLookupFailureException( message, e );
 		}
 
-		private static Uri GetJwkEndpoint( Uri authEndpoint, Guid keyId ) {
+		private static Uri GetJwkEndpoint( Uri authEndpoint, string keyId ) {
 			string authRoot = MakeSureThereIsATrailingSlash( authEndpoint );
 
-			authRoot += $"jwk/{keyId}";
+			authRoot += $"jwk/{HttpUtility.UrlEncode( keyId )}";
 
 			return new Uri( authRoot );
 		}
@@ -95,7 +96,7 @@ namespace D2L.Security.OAuth2.Keys.Default.Data {
 
 		private static string MakeSureThereIsATrailingSlash( Uri uri ) {
 			string root = uri.ToString();
-			if( root[ root.Length - 1 ] == '/' ) {
+			if( root[root.Length - 1] == '/' ) {
 				return root;
 			}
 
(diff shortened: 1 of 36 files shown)


src/D2L.Security.OAuth2/Keys/Default/Data/D2LJwksProvider.cs
```diff
@@ -77,10 +78,10 @@ Uri authEndpoint
 			return new PublicKeyLookupFailureException( message, e );
 		}
 
-		private static Uri GetJwkEndpoint( Uri authEndpoint, Guid keyId ) {
+		private static Uri GetJwkEndpoint( Uri authEndpoint, string keyId ) {
 			string authRoot = MakeSureThereIsATrailingSlash( authEndpoint );
 
-			authRoot += $"jwk/{keyId}";
+			authRoot += $"jwk/{HttpUtility.UrlEncode( keyId )}";
```
khorwood: Added UrlEncode for keyId as the source is now a string.

src/D2L.Security.OAuth2/Keys/Default/Data/D2LJwksProvider.cs
```diff
@@ -2,15 +2,16 @@
 using System.Net;
```
khorwood: Left this provider as-is to avoid forcing any changes in D2L land too soon.

src/D2L.Security.OAuth2/Keys/Default/Data/StandardJwksProvider.cs
```diff
@@ -0,0 +1,53 @@
+﻿using System;
```
khorwood: New validator for 3rd party JWKS clients

src/D2L.Security.OAuth2/Validation/AccessTokens/AccessTokenValidatorFactory.cs
```diff
@@ -52,5 +52,27 @@ Uri authEndpoint
 			return result;
 		}
 
+		/// <summary>
+		/// Creates an <see cref="IAccessTokenValidator"/> for Standard OAuth2 instance backed by a remote token signer.
+		/// </summary>
+		/// <param name="httpClient"><see cref="HttpClient"/> instance with which requests will be made. The lifecycle of the <see cref="HttpClient"/> is not managed. It will not be disposed by the validator.</param>
+		/// <param name="authEndpoint">The URI of the remote JWKS file</param>
+		/// <returns></returns>
+		public static IAccessTokenValidator CreateRemoteStandardValidator(
```
khorwood: New factory method to get a validator for 3rd party JWKS clients.
Will add a new keyset column to AuthService to allow selection based on presence.

khorwood:
*bump*

src/D2L.Security.OAuth2/Keys/Default/Data/StandardJwksProvider.cs
```diff
@@ -0,0 +1,53 @@
...
+			Uri authEndpoint
+		) {
+			m_httpClient = httpClient;
+			m_authEndpoint = authEndpoint;
+		}
+
+		string IJwksProvider.Namespace => m_authEndpoint.AbsoluteUri;
+
+		async Task<JsonWebKeySet> IJwksProvider.RequestJwkAsync( string keyId ) {
+			try {
```
omsmith: This `try/catch`ery seems unnecessary.

src/D2L.Security.OAuth2/Keys/Development/InMemoryPublicKeyDataProvider.cs
```diff
@@ -12,9 +12,9 @@ namespace D2L.Security.OAuth2.Keys.Development {
 	/// </summary>
 	[Obsolete( "Only use this in tests and for prototyping without a db" )]
 	public sealed class InMemoryPublicKeyDataProvider : IPublicKeyDataProvider {
-		private readonly ConcurrentDictionary<Guid, JsonWebKey> m_keys = new ConcurrentDictionary<Guid, JsonWebKey>();
+		private readonly ConcurrentDictionary<string, JsonWebKey> m_keys = new ConcurrentDictionary<string, JsonWebKey>();
```
omsmith: I know this is only the InMenory one, so this feedback _might_ apply elsewhere. This dictionary should probably get a comparer specified...

For Guids-as-strings, I want to say `OrdinalIgnoreCase`, but that won't work in general. So, perhaps just `Ordinal`? We don't want "equivalent" characters comparing as equal.

omsmith: Maybe we can do the `Guid -> string` first? That's mostly what this PR is, but it would be cool to not have to think about the new JwksProvider at the exact same time?

khorwood: Sure, I can build a new PR just for the `Guid -> string` conversion. Any other preferred changes in addition to the `Ordinal` suggestion?

omsmith: Don't have an immediate reference, but the public key caching, selecting a key from `JsonWebKeySet` and the "key to verify with" selection will need a look at for similar impacts.

src/D2L.Security.OAuth2/Keys/Default/Data/StandardJwksProvider.cs
khorwood: Copied from original JwksProvider, but I can review if it's necessary.

omsmith: The d2l provider is doing it's own http calls, this one is just calling through to the full-set method which already catches these things.

khorwood:
Closing in favour of an updated branch.

PR #115, opened 2019-05-13
Migrate `kid` from Guid to string for compatibility

khorwood:
Split the `Guid -> string` refactoring into it's own PR as per @omsmith 

Will rebase #114 onto this once accepted.


src/D2L.Security.OAuth2/Keys/Default/D2LSecurityToken.SecurityToken.cs
```diff
@@ -34,7 +34,7 @@ SecurityKeyIdentifierClause keyIdentifierClause
 			var clause = keyIdentifierClause as NamedKeySecurityKeyIdentifierClause;
 			return clause != null
 				&& clause.Name.Equals( OAuth2.Constants.Claims.KEY_ID, StringComparison.Ordinal )
-				&& clause.Id.Equals( KeyId.ToString(), StringComparison.OrdinalIgnoreCase );
+				&& clause.Id.Equals( KeyId, StringComparison.Ordinal );
```
khorwood: @omsmith I think this is the comparison you were referring to.

src/D2L.Security.OAuth2/Keys/Default/JsonWebKeySet.cs
```diff
@@ -39,7 +39,7 @@ internal sealed class JsonWebKeySet {
 			m_keys = ImmutableArray.Create( jsonWebKey );
 		}
 
-		public bool TryGetKey( Guid keyId, out JsonWebKey key ) {
+		public bool TryGetKey( string keyId, out JsonWebKey key ) {
 			foreach( JsonWebKey currentKey in m_keys ) {
 				if( currentKey.Id == keyId ) {
```
khorwood: I believe strict equality should be safe here.

omsmith: Yes, I do believe this does `string.Equals( a, b )` which defaults to `Ordinal`, but I'd be more comfortable if it was called out specifically.

Either `string.Equals( a, b, StringComparison.Ordinal )`, or perhaps a static "keyid comparer" utility function so we don't have to be sure to get it right everywhere?

test/D2L.Security.OAuth2.UnitTests/Keys/Default/RemotePublicKeyProviderTests.cs
```diff
@@ -17,14 +17,14 @@ internal sealed class RemotePublicKeyProviderTests {
...
-		private Guid KEY_ID;
+		private string KEY_ID;
 
 		[SetUp]
 		public void BeforeEach() {
 			m_jwksProvider = new Mock<IJwksProvider>( MockBehavior.Strict );
 			m_keyCache = new Mock<IInMemoryPublicKeyCache>( MockBehavior.Strict );
 			SRC_NAMESPACE = Guid.NewGuid().ToString();
-			KEY_ID = Guid.NewGuid();
+			KEY_ID = Guid.NewGuid().ToString();
```
omsmith: Somewhere along the line, it would be good to test non-guid key ids.

src/D2L.Security.OAuth2/Keys/Caching/InMemoryPublicKeyCache.cs
```diff
@@ -37,12 +37,12 @@ MemoryCache cache
 			);
 		}
 
-		D2LSecurityToken IInMemoryPublicKeyCache.Get( string srcNamespace, Guid keyId ) {
+		D2LSecurityToken IInMemoryPublicKeyCache.Get( string srcNamespace, string keyId ) {
 			var result = m_cache.Get( BuildCacheKey( srcNamespace, keyId ) ) as D2LSecurityToken;
```
omsmith: I believe `MemoryCache` is case-sensitive / uses HashCode, so should be okay.

test/D2L.Security.OAuth2.UnitTests/Keys/Default/RemotePublicKeyProviderTests.cs
khorwood: - [X] tests with string KeyIds

src/D2L.Security.OAuth2/Keys/Default/JsonWebKeySet.cs
khorwood: - [X] string KeyId comparison helper

src/D2L.Security.OAuth2/Keys/Caching/InMemoryPublicKeyCache.cs
khorwood: Yup, uses Ordinal: https://referencesource.microsoft.com/#System.Runtime.Caching/System/Caching/MemoryCacheKeyEqualityComparer.cs

j3parker:
What's the plan for the 7.x vs. master branch stuff?

khorwood:
@j3parker Unsure.. When we're finished I should probably catch it up, but there are some behavioural changes in there to reconcile.

omsmith:
I'll just reset master to 7.x and ditch my master changes for now.

khorwood:
Do I need to bump this to `7.4.x`?

omsmith:
Nah, we'll do it post-merge and this will put it to 8.0.0 (this repo is semvery).

I assume we might want to do a few more changes before bumping though (such as the provider changes).

khorwood:
Cool. If we're cool can we merge this, so I can put up the provider PR?

PR #117, opened 2019-05-16
Update JWKS provider for full JWKS Uri, and optional JWK Uri

khorwood:
As per discussion with @j3parker

- Will need to migrate Auth Service `public_key_endpoint` to:
  - jwks_endpoint: `x => x + /.well-known/jwks`
  - jwk_endpoint: `x => x + /jwk/`

### DIFF
--- a/src/D2L.Security.OAuth2/Keys/Default/Data/JwksProvider.cs
+++ b/src/D2L.Security.OAuth2/Keys/Default/Data/JwksProvider.cs
@@ -39,11 +39,11 @@ namespace D2L.Security.OAuth2.Keys.Default.Data {
 
 		async Task<JsonWebKeySet> IJwksProvider.RequestJwkAsync( string keyId ) {
 			var url = GetJwkEndpoint( m_jwkEndpoint, keyId );
-			try {
-				if( url == null ) {
-					return await ( this as IJwksProvider ).RequestJwksAsync().SafeAsync();
-				}
+			if( url == null ) {
+				return await ( this as IJwksProvider ).RequestJwksAsync().SafeAsync();
+			}
 
+			try {
 				using( var res = await m_httpClient.GetAsync( url ).SafeAsync() ) {
 					// This is temporary while we try to fully deprecate the
 					// JWKS route. 404 might mean the key doesn't exist (which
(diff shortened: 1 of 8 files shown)


src/D2L.Security.OAuth2/Keys/Default/Data/JwksProvider.cs
```diff
@@ -9,36 +9,41 @@
...
 		async Task<JsonWebKeySet> IJwksProvider.RequestJwkAsync( string keyId ) {
-			var url = GetJwkEndpoint( m_authEndpoint, keyId );
+			var url = GetJwkEndpoint( m_jwkEndpoint, keyId );
 			try {
+				if( url == null ) {
+					return await ( this as IJwksProvider ).RequestJwksAsync().SafeAsync();
+				}
+
 				using( var res = await m_httpClient.GetAsync( url ).SafeAsync() ) {
 					// This is temporary while we try to fully deprecate the
```
j3parker: TODO for me: re-remove this as in #110

src/D2L.Security.OAuth2/Keys/Default/Data/JwksProvider.cs
```diff
@@ -9,36 +9,41 @@
...
 			}
 		}
 
 		async Task<JsonWebKeySet> IJwksProvider.RequestJwkAsync( string keyId ) {
-			var url = GetJwkEndpoint( m_authEndpoint, keyId );
+			var url = GetJwkEndpoint( m_jwkEndpoint, keyId );
 			try {
+				if( url == null ) {
+					return await ( this as IJwksProvider ).RequestJwksAsync().SafeAsync();
+				}
```
j3parker: Does this need to be in the `try`? It looks like that method catches the same exceptions.

src/D2L.Security.OAuth2/Keys/Default/Data/JwksProvider.cs
```diff
@@ -70,7 +75,7 @@ Uri authEndpoint
 			}
 		}
 
-		string IJwksProvider.Namespace => m_authEndpoint.AbsoluteUri;
+		string IJwksProvider.Namespace => m_jwksEndpoint.AbsoluteUri;
```
j3parker: @omsmith FYI. @khorwood and I discussed this in person. There are some implicit assumptions here but seems reasonable?

j3parker:
Add some test coverage for that?
That'd be fine as a follow-up.

src/D2L.Security.OAuth2/Keys/Default/Data/JwksProvider.cs
khorwood: Probably not, no.

khorwood: - [X] move block out of try-catch

khorwood:
@j3parker Added some extra tests and moved the fallback outside of the try-catch.

PR #118, opened 2019-05-22
Revert IPublicKeyDataProvider back to Guid

khorwood:
Reverts the `IPublicKeyDataProvider` back to Guid ids as string is not
required on this interface, only used by D2L/AuthService.


src/D2L.Security.OAuth2/Keys/Default/ExpiringPublicKeyDataProvider.cs
```diff
@@ -47,7 +47,7 @@ IDateTimeProvider dateTimeProvider
 			// how GetAllAsync is implemented at the moment, but still.
 			keys = keys.ToList();
 
-			JsonWebKey key = keys.SingleOrDefault( jwk => jwk.Id == id );
+			JsonWebKey key = keys.SingleOrDefault( jwk => jwk.Id.KeyIdEquals( id.ToString() ) );
```
khorwood: This should use `jwk => new Guid( jwk.Id ) == id` instead

PR #120, opened 2019-07-10
Omsmith/fix exceptions

omsmith:


omsmith:
Future TODO: [JsonWebKeyTests](https://github.com/Brightspace/D2L.Security.OAuth2/blob/161218d57c04fe1ff184c8ef51edec6b78a4973a/test/D2L.Security.OAuth2.UnitTests/Keys/JsonWebKeyTests.cs) could do with some negative test cases.

PR #128, opened 2020-09-25
refact: DateTime => DateTimeOffset

omsmith:
~cherry-picked from `omsmith/keep`


j3parker (approved):
Excellent

src/D2L.Security.OAuth2/Keys/Default/D2LSecurityToken.cs
```diff
@@ -40,11 +40,11 @@ internal sealed partial class D2LSecurityToken : SecurityToken {
 		public string KeyId { get { return m_id; } }
 
 		public override DateTime ValidFrom {
-			get { return m_validFrom; }
+			get { return m_validFrom.UtcDateTime; }
```
omsmith, line 43: This is different then the original commit, because it was on top of the IdentityModel package update, which switched from `SecurityToken` to `SecurityKey` which didn't have these properties to override, so I was also able to change the props to `DateTimeOffset`.

Will investigate what the modern version looks like

PR #140, opened 2020-12-04
Adopt centralized package versions stuff

j3parker:


Packages.props
```diff
@@ -0,0 +1,6 @@
+<?xml version="1.0" encoding="utf-8"?>
+<Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+	<ItemGroup>
+		<PackageReference Update="Newtonsoft.Json" Version="[6.0.6]" />
```
j3parker, line 4: Just starting with this one for now, I want a `-preview` build of stuff with 7.0.1 (follow-up PR)

j3parker: Oh, it's all or nothing eh

Packages.props
```diff
@@ -0,0 +1,21 @@
+<?xml version="1.0" encoding="utf-8"?>
+<Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+	<ItemGroup>
+    	<PackageReference Update="BenchmarkDotNet" Version="[0.12.1]" />
```
j3parker, line 4: Probably breaks dependabot :( I like how we do it in the LMS for now.

src/D2L.Security.OAuth2.WebApi/D2L.Security.OAuth2.WebApi.csproj
```diff
@@ -1,4 +1,5 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
+	<Sdk Name="Microsoft.Build.CentralPackageVersions" Version="2.0.1" />
```
JeffAshton, line 2: You can also centralize the version of this with `global.json`:

https://search.d2l.dev/xref/lms/global.json?r=66ab99aa

Packages.props
JeffAshton: Visual studio will also do updates by setting the version in the projects, so people just need ot keep an eye out in reviews.  But the centralized is much nicer.

JeffAshton (approved):

src/D2L.Security.OAuth2.WebApi/D2L.Security.OAuth2.WebApi.csproj
j3parker: Oh nice, will do.

PR #162, opened 2021-04-12
Disabling signature provider caching

JeffAshton:


src/D2L.Security.OAuth2/Keys/Default/D2LSecurityToken.cs
```diff
@@ -46,7 +46,9 @@ internal sealed partial class D2LSecurityToken : SecurityKey {
 				this,
 				GetSignatureAlgorithm()
 			) {
-				CryptoProviderFactory = new D2LCryptoProviderFactory()
+				CryptoProviderFactory = new D2LCryptoProviderFactory {
+					CacheSignatureProviders = false
```
j3parker, line 50: https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/wiki/Caching-in-Microsoft.IdentityModel
https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/pull/1605

j3parker (approved):

j3parker:
@omsmith  ^ ?

omsmith (approved):
I would prefer this gets set inside `D2LCryptoProviderFactory`, but otherwise looks good.

omsmith:
This doesn't prevent the cache itself from getting made, so it's still going to leak tasks.

JeffAshton:
> This doesn't prevent the cache itself from getting made, so it's still going to leak tasks.

well that sucks then

PR #166, opened 2021-04-15
Make bare access token validation public

j3parker:


src/D2L.Security.OAuth2/Validation/Request/IRequestAuthenticator.cs
```diff
@@ -1,13 +1,8 @@
 ﻿using System.Net.Http;
 using System.Threading.Tasks;
 using D2L.Security.OAuth2.Principal;
-using D2L.Security.OAuth2.Validation.AccessTokens;
 
 namespace D2L.Security.OAuth2.Validation.Request {
-	/// <summary>
-	/// An abstraction for authenticating access tokens that works at the request level
-	/// rather than the token level (see <see cref="IAccessTokenValidator"/>)
-	/// </summary>
```
j3parker, line 10: IAccessTokenValidator is public, but the D2LPrincipal construction isn't. I didn't think this comment was completely accurate before and this change made it totally inaccurate.

src/D2L.Security.OAuth2/Validation/Request/RequestAuthenticator.cs
```diff
@@ -19,10 +19,10 @@ HttpRequestMessage request
 		) {
 			string bearerToken = request.GetBearerTokenValue();
 
-			return AuthenticateHelper( bearerToken );
+			return AuthenticateAsync( bearerToken );
 		}
 
-		private async Task<ID2LPrincipal> AuthenticateHelper(
+		public async Task<ID2LPrincipal> AuthenticateAsync(
```
j3parker, line 25: .NET 5 has `HttpRequest` from [a NuGet package](https://www.nuget.org/packages/Microsoft.AspNetCore.Http/).

I could add an overload, but then we would need to dep on this package for the .NET 5 build. This sucks because the dependency is maybe not necessary in some cases (imagine a Lambda function taking tokens in its request body for whatever reason). It mostly sucks because its another dependency to rev in this DLL which makes upgrades more annoying.

Previously we had a separate ASP.NET integration DLL. I like that pattern but stuff is a bit whack right now in the transition.

Since `IAccessTokenValidator` is public I don't see any reason to not make this one public.

omsmith (approved):

PR #191, opened 2021-08-19
Allow private keys to be stored and rotated externally.

j3parker:
External storage (vs. in-memory) will work better for scale-up/scale-down activity. This will enable us to scale more aggressively, including use-cases like AWS Lambda.

It also moves rotation out-of-line of requests, which will avoid CPU+latency spikes.

Fetching the latest private key will still be done on-demand, but there is the option to do it proactively out-of-band. This will allow us to consolidate the fetch in web-server workloads and further hide latency from individual requests.

Web server workloads will be able to proactively fetch a key on boot before being placed into load.

### DIFF
--- a/src/D2L.Security.OAuth2/IsExternalInit.net4x.cs
+++ b/src/D2L.Security.OAuth2/IsExternalInit.net4x.cs
@@ -1,0 +1,16 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+// From: https://github.com/dotnet/runtime/commit/6072e4d3a7a2a1493f514cdf4be75a3d56580e84#diff-afc57cc6e9c659ebf491bc78d06cdb5b13d5a049122d8a41f744a515f493663b
+
+using System.ComponentModel;
+
+namespace System.Runtime.CompilerServices {
+	/// <summary>
+	/// Reserved to be used by the compiler for tracking metadata.
+	/// This class should not be used by developers in source code.
+	/// </summary>
+	[EditorBrowsable( EditorBrowsableState.Never )]
+	internal static class IsExternalInit {
+	}
+}
--- a/src/D2L.Security.OAuth2/Keys/IKeyManagementService.cs
+++ b/src/D2L.Security.OAuth2/Keys/IKeyManagementService.cs
@@ -1,0 +1,24 @@
+﻿using System;
+using System.Threading.Tasks;
+
+namespace D2L.Security.OAuth2.Keys {
+	public interface IKeyManagementService {
+		/// <summary>
+		/// Generates a new key (storing it's public and private keys) if the
+		/// existing ones are getting near their expiry.
+		///
+		/// You would call this in a background job on a regular + frequent
+		/// cadence.
+		/// </summary>
+		Task GenerateNewKeyIfNeededAsync();
+
+		/// <summary>
+		/// Causes the key manager to go looking for a new private key from the
+		/// key store.
+		/// </summary>
+		/// <returns>
+		/// The amount of time to wait before calling this again.
+		/// </returns>
+		Task<TimeSpan> RefreshKeyAsync();
+	}
+}
--- a/src/D2L.Security.OAuth2/Keys/IPrivateKeyDataProvider.cs
+++ b/src/D2L.Security.OAuth2/Keys/IPrivateKeyDataProvider.cs
@@ -1,0 +1,24 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Threading.Tasks;
+
+namespace D2L.Security.OAuth2.Keys {
+	/// <summary>
+	/// An externally-implemented storage for private keys.
+	/// </summary>
+	public interface IPrivateKeyDataProvider {
+		/// <summary>
+		/// Save a private key.
+		/// </summary>
+		Task SaveAsync( PrivateKeyData keyData );
+
+		/// <summary>
+		/// Gets all non-expiring private keys.
+		/// </summary>
+		/// <param name="validUntilAtLeast">The earliest possible expiry time toreturn (should be in the future.)</param>
+		/// <returns>All non-expired/expiring private keys</returns>
+		Task<IEnumerable<PrivateKeyData>> GetAllAsync(
+			DateTimeOffset validUntilAtLeast
+		);
+	}
+}
--- a/src/D2L.Security.OAuth2/Keys/KeyManagementService.cs
+++ b/src/D2L.Security.OAuth2/Keys/KeyManagementService.cs
@@ -36,42 +36,46 @@ namespace D2L.Security.OAuth2.Keys {
 			if ( current == null || current.ValidTo <= now ) {
 				// Slow path: RefreshKeyAsync() wasn't called on boot and/or it
 				// isn't being called in a background job.
-				await RefreshKeyAsync( current, now )
+				await RefreshKeyAsync( now )
 					.ConfigureAwait( false );
 
 				current = Volatile.Read( ref m_current );
+			}
+
+			if( current == null ) {
+				return null;
 			}
 
 			return current.Ref();
 		}
 
 		async Task<TimeSpan> IKeyManagementService.RefreshKeyAsync() {
-			var current = Volatile.Read( ref m_current );
-
 			var now = m_clock.UtcNow;
 
-			await RefreshKeyAsync(
-				current,
-				now
-			).ConfigureAwait( false );
+			await RefreshKeyAsync( now )
+				.ConfigureAwait( false );
 
-			current = Volatile.Read( ref m_current );
+			var current = Volatile.Read( ref m_current );
 
 			if( current == null || now > current.ValidTo ) {
 				// If the key is expired or doesn't exist, retry quickly.
 				return TimeSpan.FromSeconds( 10 );
 			}
 
-			var expectedNextRotation = current.ValidTo - m_config.KeyRotationBuffer;
+			// A new key will get generated some time before the current key
+			// expires, but will only become usable some time after that.
+			var expectedTimeOfNewUsableKey = current.ValidTo
+				- m_config.KeyRotationBuffer
+				+ m_config.KeyTimeUntilUse;
 
-			if( now > current.ValidTo - m_config.KeyRotationBuffer ) {
+			if( now > expectedTimeOfNewUsableKey ) {
 				// If we would have expected a new key by now, retry again in a
 				// bit. This code branch supports configuration changes mostly.
 				return TimeSpan.FromMinutes( 1 );
 			} else {
 				// Otherwise use that but with a little buffer for key
 				// generation time/imprecisely scheduled cron jobs.
-				return expectedNextRotation.AddMinutes( 1 ) - now;
+				return expectedTimeOfNewUsableKey.AddMinutes( 1 ) - now;
 			}
 		}
 
@@ -83,7 +87,7 @@ namespace D2L.Security.OAuth2.Keys {
 			).ConfigureAwait( false );
 
 			foreach( var key in keys ) {
-				if( now < key.ExpiresAt - m_config.KeyRotationBuffer ) {
+				if( !key.WouldPreferToRotate( now, m_config.KeyRotationBuffer ) ) {
 					// Found a suitable key, so we don't need to generate a new one.
 					return;
 				}
@@ -128,15 +132,12 @@ namespace D2L.Security.OAuth2.Keys {
 			).ConfigureAwait( false );
 		}
 
-		private async Task RefreshKeyAsync(
-			D2LSecurityToken current,
-			DateTimeOffset now
-		) {
+		private async Task RefreshKeyAsync( DateTimeOffset now ) {
 			var keys = await m_privateKeys.GetAllAsync(
 				validUntilAtLeast: now
 			).ConfigureAwait( false );
 
-			var best = ChooseKey( keys, now ).Ref();
+			var best = ChooseKey( keys, now )?.Ref();
 
 			if( best == null ) {
 				// If we didn't find anything in the database, continue with
@@ -161,20 +162,30 @@ namespace D2L.Security.OAuth2.Keys {
 					continue;
 				}
 
+				// The data provider should filter expired keys, but we don't
+				// rely on that just in case.
+				if( key.IsExpired( now ) ) {
+					continue;
+				}
+
 				// Use any non-expired key if it's the only one
 				if( candidate == null ) {
 					candidate = key;
 					continue;
 				}
 
-				if( candidate.NotBefore > now && key.NotBefore <= now ) {
-					// If we can switch to a key past its NotBefore, do that.
+				var candidateIsPastNotBefore = candidate.IsPastNotBefore( now );
+				var keyIsPastNotBefore = key.IsPastNotBefore( now );
+
+				if( !candidateIsPastNotBefore && keyIsPastNotBefore ) {
+					// Prefer keys past their "NotBefore" date because they are
+					// guarunteed to not be excluded from caches.
 					candidate = key;
-				} else if( candidate.NotBefore > now && candidate.CreatedAt > key.CreatedAt ) {
-					// When comparing two keys that are not past their respective
-					// NotBefore points, prefer the oldest one.
+				} else if( !candidateIsPastNotBefore && key.CreatedAt < candidate.CreatedAt ) {
+					// When comparing two keys that are not past their NotBefore
+					// points, prefer the oldest one (same cache rationale.)
 					candidate = key;
-				} else if ( candidate.NotBefore <= now && key.NotBefore <= now && candidate.CreatedAt < key.CreatedAt ) {
+				} else if( candidateIsPastNotBefore && keyIsPastNotBefore && candidate.CreatedAt < key.CreatedAt ) {
 					// If we have two keys past their NotBefore date, pick the
 					// one that was created most recently.
 					candidate = key;
--- a/src/D2L.Security.OAuth2/Keys/KeyManagementService.cs
+++ b/src/D2L.Security.OAuth2/Keys/KeyManagementService.cs
@@ -96,9 +96,10 @@ namespace D2L.Security.OAuth2.Keys {
 			var keyId = Guid.NewGuid();
 			var keyIdStr = keyId.ToString();
 
-			using var csp = new RSACryptoServiceProvider {
+			using var csp = new RSACryptoServiceProvider( Constants.GENERATED_RSA_KEY_SIZE ) {
 				PersistKeyInCsp = false
 			};
+
 
 			var pub = csp.ExportParameters(
 				includePrivateParameters: false
--- a/src/D2L.Security.OAuth2/Keys/KeyManagementService.cs
+++ b/src/D2L.Security.OAuth2/Keys/KeyManagementService.cs
@@ -1,0 +1,207 @@
+﻿using D2L.Security.OAuth2.Keys.Default;
+using D2L.Security.OAuth2.Utilities;
+using Microsoft.IdentityModel.Tokens;
+using System;
+using System.Collections.Generic;
+using System.Security.Cryptography;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace D2L.Security.OAuth2.Keys {
+	internal sealed class KeyManagementService : IKeyManagementService, IPrivateKeyProvider, IDisposable {
+		private readonly IPublicKeyDataProvider m_publicKeys;
+		private readonly IPrivateKeyDataProvider m_privateKeys;
+		private readonly IDateTimeProvider m_clock;
+		private readonly OAuth2Configuration m_config;
+
+		private D2LSecurityToken m_current = null;
+
+		public KeyManagementService(
+			IPublicKeyDataProvider publicKeys,
+			IPrivateKeyDataProvider privateKeys,
+			IDateTimeProvider clock,
+			OAuth2Configuration config
+		) {
+			m_publicKeys = publicKeys;
+			m_privateKeys = privateKeys;
+			m_clock = clock;
+			m_config = config;
+		}
+
+		async Task<D2LSecurityToken> IPrivateKeyProvider.GetSigningCredentialsAsync() {
+			var current = Volatile.Read( ref m_current );
+
+			var now = m_clock.UtcNow;
+
+			if ( current == null || current.ValidTo <= now ) {
+				// Slow path: RefreshKeyAsync() wasn't called on boot and/or it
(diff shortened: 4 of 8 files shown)


j3parker:
What's missing:

* Tests
* ECDSA support
* Refactoring the in-memory stuff to be built on top of this

The first of those is mandatory to merge this, the other two I'd be good to do in a follow-up (like immediately after, not "in the future")

Getting this PR up without tests just to validate the approach.

src/D2L.Security.OAuth2/IsExternalInit.net4x.cs
```diff
@@ -0,0 +1,16 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+// From: https://github.com/dotnet/runtime/commit/6072e4d3a7a2a1493f514cdf4be75a3d56580e84#diff-afc57cc6e9c659ebf491bc78d06cdb5b13d5a049122d8a41f744a515f493663b
+
+using System.ComponentModel;
+
+namespace System.Runtime.CompilerServices {
```
j3parker, line 8: private polyfill, only for .NET Framework (see path -- `.net4x.cs` and the magic we do in csproj for multi-platform.)

j3parker: This is for `init;` in the `OAuth2Configuration` file -- I could drop it if it rankles.

src/D2L.Security.OAuth2/Keys/IKeyManagementService.cs
```diff
@@ -0,0 +1,24 @@
...
+		/// <summary>
+		/// Causes the key manager to go looking for a new private key from the
+		/// key store.
+		/// </summary>
+		/// <returns>
+		/// The amount of time to wait before calling this again.
+		/// </returns>
+		Task<TimeSpan> RefreshKeyAsync();
+	}
+}
```
j3parker, line 24: This is the public interface. Very high level:

ASP.NET will call `RefreshKeyAsync` in a background worker, using the returned `TimeSpan` to self-throttle. The auth service will use `GenerateNewKeyIfNeededAsync` in a CloudWatch cron-job/Lambda.

Eventually I'd like to do something similar in the LMS.

src/D2L.Security.OAuth2/Keys/IPrivateKeyDataProvider.cs
```diff
@@ -0,0 +1,24 @@
...
+
+namespace D2L.Security.OAuth2.Keys {
+	/// <summary>
+	/// An externally-implemented storage for private keys.
+	/// </summary>
+	public interface IPrivateKeyDataProvider {
+		/// <summary>
+		/// Save a private key.
+		/// </summary>
+		Task SaveAsync( PrivateKeyData keyData );
```
j3parker, line 13: We talked about not including this. It's back because when playing with where keygen should go it's nice if `KeyManagementService` is responsible for both generating and storing the key (to both places.)

We could theoretically have one `IKeyDataProvider` interface rather than the separation, but I think that'd be something to consider if we totally redesigned the library.

src/D2L.Security.OAuth2/Keys/KeyManagementService.cs
```diff
@@ -0,0 +1,207 @@
...
+		}
+
+		async Task<D2LSecurityToken> IPrivateKeyProvider.GetSigningCredentialsAsync() {
+			var current = Volatile.Read( ref m_current );
+
+			var now = m_clock.UtcNow;
+
+			if ( current == null || current.ValidTo <= now ) {
+				// Slow path: RefreshKeyAsync() wasn't called on boot and/or it
+				// isn't being called in a background job.
```
j3parker, line 38: We want to never hit the inside of this branch in ASP.NET (even on boot.)

src/D2L.Security.OAuth2/Keys/KeyManagementService.cs
```diff
@@ -0,0 +1,207 @@
...
+
+			var expectedNextRotation = current.ValidTo - m_config.KeyRotationBuffer;
+
+			if( now > expectedNextRotation ) {
+				// If that's in the past, use a short retry window.
+				return TimeSpan.FromSeconds( 10 );
+			} else {
+				// Otherwise use that but with a little buffer for key
+				// generation time/imprecisely scheduled cron jobs.
+				return expectedNextRotation.AddMinutes( 1 ) - now;
```
j3parker, line 68: Both of these numbers assume a little bit about the numbers in the config. `OAuth2Configuration` sets some defaults and minimums for these numbers so that this is "obviously" safe -- see it's class further down.

src/D2L.Security.OAuth2/Keys/KeyManagementService.cs
```diff
@@ -0,0 +1,207 @@
...
+						id: keyIdStr,
+						kind: PrivateKeyData.KeyKinds.Rsa,
+						data: priv,
+						createdAt: now,
+						notBefore: now + m_config.KeyTimeUntilUse,
+						expiresAt: expiresAt
+					)
+				)
+			).ConfigureAwait( false );
+		}
```
j3parker, line 123: This probably needs a bit of organization once we add ECDSA support, at least.

src/D2L.Security.OAuth2/Keys/PrivateKeyData.cs
```diff
@@ -0,0 +1,62 @@
...
+		/// <summary>
+		/// The key should not be used to sign things before this point. The
+		/// gap between CreatedAt and NotBefore allows for jwks caches to
+		/// expire and gives opportunity for people to fetch our new key in the
+		/// background.
+		//
+		/// If there are no other keys available in an emergency it is okay
+		/// to sign messages with a key that has not reached NotBefore.
+		/// </summary>
+		public DateTimeOffset NotBefore { get; }
```
j3parker, line 55: I got rid of `BestBefore` and jitter + debouncing, for simplicity.

I think there is still value in that, but communicating all that information back to the client (e.g. the auth service) is a little tricky/would require some more API design. I don't think, given the things that will use this library and how they're implemented, we don't need to worry about thundering herd problems anymore (given what things we plan to shift to background jobs (i.e. shrinking the herd) + the number of processes we're talking about + the database where we are storing things).

src/D2L.Security.OAuth2/OAuth2Configuration.cs
```diff
@@ -0,0 +1,59 @@
...
+		/// key will become invalid before it expires.
+		/// </summary>
+		public TimeSpan KeyRotationBuffer { get; init; }
+			= TimeSpan.FromDays( 7 );
+
+		/// <summary>
+		/// The total lifetime keys we generate.
+		/// </summary>
+		public TimeSpan KeyLifetime { get; init; }
+			= TimeSpan.FromDays( 30 );
```
j3parker, line 23: These defaults are what I was thinking of for the auth service and eventually the LMS.

src/D2L.Security.OAuth2/OAuth2Configuration.cs
```diff
@@ -0,0 +1,59 @@
...
+			}
+
+			if( KeyLifetime.TotalDays > 365 ) {
+				throw new Exception( "KeyLifetime is too long" );
+			}
+
+			if( (KeyLifetime - KeyRotationBuffer - KeyTimeUntilUse).TotalHours < 1 ) {
+				throw new Exception( "KeyLifetime should be long enough to avoid rotations for at least an hour" );
+			}
+		}
```
j3parker, line 57: Didn't bother with good error handling because I don't expect we'll be tweaking this a bunch right away.
This would cause things to fail to boot which is nasty but probably right.

src/D2L.Security.OAuth2/OAuth2Factory.cs
```diff
@@ -0,0 +1,38 @@
...
+		public static void Create(
+			IPublicKeyDataProvider publicKeyDataProvider,
+			IPrivateKeyDataProvider privateKeyDataProvider,
+
+			out IKeyManagementService keyManagementService,
+			out ITokenSigner tokenSigner,
+
+			OAuth2Configuration config = null
+		) {
+			config = config ?? new OAuth2Configuration();
```
j3parker, line 20: I plan to rely on this in practice for now.

src/D2L.Security.OAuth2/OAuth2Factory.cs
```diff
@@ -0,0 +1,38 @@
...
+using D2L.Security.OAuth2.Keys.Default;
+using D2L.Security.OAuth2.Utilities;
+
+namespace D2L.Security.OAuth2 {
+	public static class OAuth2Factory {
+		/// <summary>
+		/// Create an IKeyManagmentService and ITokenSigner given key storage
+		/// and configuration.
+		/// </summary>
+		public static void Create(
```
j3parker, line 11: This is a replacement for e.g. `RsaTokenSignerFactory`.

We could have an overload that doesn't take `IPrivateKeyDataProvider` and uses in-memory storage, and then start trimming away some of the old stuff.

src/D2L.Security.OAuth2/Keys/PrivateKeyData.cs
```diff
@@ -0,0 +1,62 @@
+﻿using System;
+
+namespace D2L.Security.OAuth2.Keys {
+	public sealed class PrivateKeyData {
+		internal static class KeyKinds {
+			public const string Rsa = "rsa";
+			public const string Ecdsa = "ecdsa";
+		}
+
+		public PrivateKeyData(
```
j3parker, line 10: This has to be `public` so that it can be deserialized by the implementer of `IPrivateKeyDataProvider`.

src/D2L.Security.OAuth2/Keys/KeyManagementService.cs
```diff
@@ -0,0 +1,207 @@
...
+			var keys = await m_privateKeys.GetAllAsync(
+				validUntilAtLeast: now
+			).ConfigureAwait( false );
+
+			foreach( var key in keys ) {
+				if( now < key.ExpiresAt - m_config.KeyRotationBuffer ) {
+					// This key is fine, do nothing.
+					return;
+				}
+			}
```
omsmith, line 84: Believe this will never generate a second key (for priming)?

src/D2L.Security.OAuth2/Keys/KeyManagementService.cs
```diff
@@ -0,0 +1,207 @@
...
+						pub
+					)
+				),
+				m_privateKeys.SaveAsync(
+					new PrivateKeyData(
+						id: keyIdStr,
+						kind: PrivateKeyData.KeyKinds.Rsa,
+						data: priv,
+						createdAt: now,
+						notBefore: now + m_config.KeyTimeUntilUse,
```
omsmith, line 118: If we didn't have a key / are recovering, I believe we won't want the `+ KeyTimeUntilUse`? (i.e. we want to start using it right away)

src/D2L.Security.OAuth2/Keys/KeyManagementService.cs
```diff
@@ -0,0 +1,207 @@
...
+					candidate = key;
+				} else if( candidate.NotBefore > now && candidate.CreatedAt > key.CreatedAt ) {
+					// When comparing two keys that are not past their respective
+					// NotBefore points, prefer the oldest one.
+					candidate = key;
+				} else if ( candidate.NotBefore <= now && key.NotBefore <= now && candidate.ExpiresAt < key.ExpiresAt ) {
+					// If we have two keys past their NotBefore date, pick the
+					// one that expires the latest.
+					candidate = key;
+				}
```
omsmith, line 173: Should this be created most recently?

Generate Key A, 3 month expiry
Adjust config, setting expiry to 1 week.
Generate Key B
Generate Key C
Generate Key D

B, C, D will all be ignored in favour of using A

src/D2L.Security.OAuth2/Keys/KeyManagementService.cs
```diff
@@ -0,0 +1,207 @@
...
+			);
+		}
+
+		private readonly object m_disposeLock = new();
+
+		public void Dispose() {
+			lock( m_disposeLock ) {
+				m_current?.Dispose();
+			}
+		}
```
omsmith, line 205: If we're disposable, should have disposed checks in methods I think?

src/D2L.Security.OAuth2/Keys/KeyManagementService.cs
```diff
@@ -0,0 +1,207 @@
...
+			D2LSecurityToken current,
+			DateTimeOffset now
+		) {
+			var keys = await m_privateKeys.GetAllAsync(
+				validUntilAtLeast: now
+			).ConfigureAwait( continueOnCapturedContext: false );
+
+			var best = ChooseKey( keys, now ).Ref();
+
+			var loser = Interlocked.CompareExchange(
```
omsmith, line 135: `CompareExchange` doesn't return the loser. It always returns the original value of `m_current`

src/D2L.Security.OAuth2/Keys/KeyManagementService.cs
j3parker: It will generate a new key if there are none that are at least `m_config.KeyRotationBuffer` (default one week) before their expiry.

src/D2L.Security.OAuth2/Keys/KeyManagementService.cs
j3parker: That's handled more generally by `ChooseKey`, rather than at generation time.

src/D2L.Security.OAuth2/Keys/KeyManagementService.cs
j3parker: I believe the spirit (but not the law) is that you shouldn't use an object after it's disposed, other than possibly calling Dispose.
Sometimes there are `if( disposed )` checks within `Dispose` impls to support it being called twice, but here we're fine.

I'm not really sure what I was thinking with this lock though. By this logic `Dispose` needs to be threadsafe but the lock wasn't necessary for that here. So I think I'll go the other direction and remove it.

src/D2L.Security.OAuth2/Keys/KeyManagementService.cs
j3parker: Good call -- good behaviour to check with tests.

Changed it to Exchange which changes the intended behaviour to always using the "last" key we pick under contention which is fine. The original thought with CompareExchange (modulo correct disposal) was to avoid changing our mind and signing tokens with a different key suddenly, but (1) if there were a different choice of key so suddenly that is ... worrying (2) we have more than one server, so different choices would be made anyway.

(Every time we use a private key we force people to fetch the corresponding public key.)

src/D2L.Security.OAuth2/Keys/KeyManagementService.cs
j3parker: 🤷‍♂️ I can see the case for that but I have no feelings (other than doing something reasonable and predictable.) I'll change it to that.

test/D2L.Security.OAuth2.UnitTests/Keys/Default/KeyManagementServiceTests.cs
```diff
@@ -0,0 +1,377 @@
...
+				.Verifiable();
+
+			m_privateKeys.Setup( priv => priv.SaveAsync( It.IsAny<PrivateKeyData>() ) )
+				.Returns( Task.CompletedTask )
+				.Verifiable();
+		}
+
+		private void VerifyNewKeyWasSaved() {
+			m_publicKeys.VerifyAll();
+			m_privateKeys.VerifyAll();
```
j3parker, line 275: In case you don't remember verifiable mocks (they suck), if those `Setup`s above aren't hit at least once this will throw (confirmed manually.) If they are called multiple times it will pass though.

I'm kind of thinking that in an ideal world this library would have one interface for callers to implement (for key storage) and things would look a bit different.

Trying to strike a balancer here between making tests that test something and not getting too into the implementation details. 🤷

test/D2L.Security.OAuth2.UnitTests/Keys/Default/KeyManagementServiceTests.cs
```diff
@@ -0,0 +1,377 @@
...
+			// have happened
+			Assert.AreEqual(
+				// Time until we generate a new key
+				ALittleBit
+				// Time until that key is active
+				+ KeyConfig.KeyTimeUntilUse
+				// Wiggle room
+				+ TimeSpan.FromMinutes( 1 ),
+				actualDt
+			);
```
j3parker, line 159: I only added this assert to one test (for the happy cases) because it's really fiddly, but it felt important.
I tested `actualDt` for the cases where we aren't doing this specific computation.

src/D2L.Security.OAuth2/Keys/KeyManagementService.cs
```diff
@@ -0,0 +1,226 @@
...
+					var csp = new RSACryptoServiceProvider {
+						PersistKeyInCsp = false
+					};
+
+					csp.ImportCspBlob( candidate.Data );
+
+					var key = new RsaSecurityKey( csp );
+
+					return new(key, csp);
+				},
```
omsmith, line 210: Believe you had ecdsa as a TODO. Do you want me to fully review what's in place now, and then that'll get added before landing?

j3parker: 🤷 I was going to do it as a follow-up.
(It isn't something we currently use -- we do use the validation side of things, but we never generate those sort of keys.)

j3parker: (Supporting it here isn't tricky, it just seemed like a distraction from this PR.)

src/D2L.Security.OAuth2/Keys/KeyManagementService.cs
```diff
@@ -0,0 +1,226 @@
...
+					return;
+				}
+			}
+
+			var keyId = Guid.NewGuid();
+			var keyIdStr = keyId.ToString();
+
+			using var csp = new RSACryptoServiceProvider {
+				PersistKeyInCsp = false
+			};
```
omsmith, line 101:
```suggestion
			using var csp = new RSACryptoServiceProvider( Constants.GENERATED_RSA_KEY_SIZE ) {
				PersistKeyInCsp = false
			};

```

src/D2L.Security.OAuth2/OAuth2Configuration.cs
```diff
@@ -0,0 +1,59 @@
...
+				throw new Exception( "KeyLifetime is too short" );
+			}
+
+			if( KeyLifetime.TotalDays > 365 ) {
+				throw new Exception( "KeyLifetime is too long" );
+			}
+
+			if( (KeyLifetime - KeyRotationBuffer - KeyTimeUntilUse).TotalHours < 1 ) {
+				throw new Exception( "KeyLifetime should be long enough to avoid rotations for at least an hour" );
+			}
```
omsmith, line 56: In the node stuff I found it a bit easier to configure any of the phases as non-overlapping (i.e. don't have an overall keylifetime you can mess up).

* Staging / TimeBeforeUse
* Use
* MaxTokenLifetime (x2 maybe) / RotationBuffer

Then just Lifetime = Staging + Use + MaxTokenLifetime

This is easier to get 30 days as a nice round number though.

Doesn't matter to me

src/D2L.Security.OAuth2/OAuth2Configuration.cs
```diff
@@ -0,0 +1,59 @@
...
+		/// make reasoning about details around rotation easier.
+		/// </summary>
+		internal void CheckSanity() {
+			if( KeyTimeUntilUse.TotalHours < 1 ) {
+				throw new Exception( "KeyTimeUntilUse is too short" );
+			}
+
+			if( KeyTimeUntilUse >= KeyLifetime ) {
+				throw new Exception( "KeyTimeUntilUse is too long" );
+			}
```
omsmith, line 36: Redundant given the final check.

src/D2L.Security.OAuth2/OAuth2Factory.cs
```diff
@@ -0,0 +1,38 @@
...
+			IPublicKeyDataProvider publicKeyDataProvider,
+			IPrivateKeyDataProvider privateKeyDataProvider,
+
+			out IKeyManagementService keyManagementService,
+			out ITokenSigner tokenSigner,
+
+			OAuth2Configuration config = null
```
omsmith, line 18: Key types would go in `OAuth2Configuration`?

test/D2L.Security.OAuth2.UnitTests/Keys/Default/KeyManagementServiceTests.cs
```diff
@@ -0,0 +1,377 @@
...
+			Assert.IsNull( creds );
+		}
+
+		[Test]
+		public async Task GenerateNewKeyIfNeededAsync_HasActiveKey_DoesNothing() {
+			var key = CreateActiveKey();
+
+			UsePrivateKeys( key );
+
+			await m_kms.RefreshKeyAsync();
```
omsmith, line 217:
```suggestion
			await m_kms.GenerateNewKeyIfNeededAsync();
```

test/D2L.Security.OAuth2.UnitTests/Keys/Default/KeyManagementServiceTests.cs
```diff
@@ -0,0 +1,377 @@
...
+		}
+
+		[Test]
+		public async Task GetSigningCredentialsAsync_TwoActiveKeys_PrefersNewest() {
+			var oldActive = CreateActiveKey();
+			var newActive = CreateNewActiveKey();
+
+			UsePrivateKeys( oldActive, newActive );
+
+			await m_kms.RefreshKeyAsync();
```
omsmith, line 129: nit: I think this test aligns more with the `OnBoot` tests, allowing `GetSigningCredentialsAsync` to call `RefreshAsync` itself, given we're going through it?

test/D2L.Security.OAuth2.UnitTests/Keys/Default/KeyManagementServiceTests.cs
j3parker: Yikes -- thanks.

src/D2L.Security.OAuth2/OAuth2Factory.cs
j3parker: That makes sense to me.

src/D2L.Security.OAuth2/OAuth2Configuration.cs
j3parker: Yeah, it exists to have a more specific error message 🤷

omsmith (approved):

PR #201, opened 2021-11-10
Fix package version constraints and switch to Directory.Package.props

j3parker:


Directory.Packages.props
```diff
@@ -0,0 +1,22 @@
...
+		<PackageReference Include="D2L.CodeStyle.Analyzers" Version="0.142.0" />
+		<PackageReference Include="D2L.CodeStyle.Annotations" Version="[0.28.0)" />
+		<PackageReference Include="D2L.Services.Core.Exceptions" Version="[2.0.1.15)" />
+		<PackageReference Include="FluentAssertions" Version="6.2.0" />
+		<PackageReference Include="Microsoft.AspNet.WebApi.Core" Version="[5.2.7]" />
+		<PackageReference Include="Microsoft.AspNet.WebApi.OwinSelfHost" Version="[5.2.7]" />
+		<PackageReference Include="Microsoft.Extensions.Caching.Memory" Version="[5.0.0)" />
+		<PackageReference Include="Microsoft.NET.Test.Sdk" Version="17.0.0" />
+		<PackageReference Include="Moq" Version="4.16.1" />
+		<PackageReference Include="Newtonsoft.Json" Version="[6,7)" />
```
j3parker, line 13: We could probably do with `[6)` but we haven't tested that.

Directory.Packages.props
```diff
@@ -0,0 +1,22 @@
+﻿<Project>
+	<ItemGroup>
+    	<PackageVersion  Include="BenchmarkDotNet" Version="0.13.1" />
+		<PackageVersion  Include="D2L.CodeStyle.Analyzers" Version="0.142.0" />
+		<PackageVersion  Include="D2L.CodeStyle.Annotations" Version="0.28.0" />
+		<PackageVersion  Include="D2L.Services.Core.Exceptions" Version="2.0.1.15" />
+		<PackageVersion  Include="FluentAssertions" Version="6.2.0" />
+		<PackageVersion  Include="Microsoft.AspNet.WebApi.Core" Version="[5.2.7]" />
+		<PackageVersion  Include="Microsoft.AspNet.WebApi.OwinSelfHost" Version="[5.2.7]" />
```
mthjones, line 9: How come these two are pinned? I doubt there will be any future releases since it was last updated in 2018, but still feels a little strange.

Directory.Packages.props
```diff
@@ -0,0 +1,22 @@
...
+		<PackageVersion  Include="Microsoft.Extensions.Caching.Memory" Version="5.0.0" />
+		<PackageVersion  Include="Microsoft.NET.Test.Sdk" Version="17.0.0" />
+		<PackageVersion  Include="Moq" Version="4.16.1" />
+		<PackageVersion  Include="Newtonsoft.Json" Version="[6.0.8,7)" />
```
omsmith, line 13: `[6.0.8,7)`?

mthjones, line 13: `6.*` may be slightly clearer, but no preference.

mthjones (approved):

Directory.Packages.props
j3parker: `<= 6.0.8 /\ < 7`

Directory.Packages.props
j3parker: Uh good question, those are actually the latest.

I got an unrelated build error and added this but it presumably did nothing. I was fine doing that because yikes @ those packages in general (the D2L.Security.OAuth2.WebApi package is effectively "deprecated" at this point, but there is still one user...)

omsmith (approved):

omsmith (approved):

Directory.Packages.props
j3parker: Yeah, I'm fine with that but don't really care, mostly I'm nervous about upgrading past 6 when we haven't vetted it (this concern is specific to newtonsoft)

j3parker: Also, TBF, I doubt there will be another 6.0.x version at this point :D

PR #202, opened 2021-11-10
Weaken version constraints

j3parker:


Packages.props
```diff
@@ -1,23 +1,23 @@
...
-		<PackageReference Update="D2L.CodeStyle.Annotations" Version="[0.28.0]" />
-		<PackageReference Update="D2L.Services.Core.Exceptions" Version="[2.0.1.15]" />
-		<PackageReference Update="FluentAssertions" Version="[6.2.0]" />
+    	<PackageReference Update="BenchmarkDotNet" Version="0.13.1" />
+		<PackageReference Update="D2L.CodeStyle.Analyzers" Version="0.142.0" />
+		<PackageReference Update="D2L.CodeStyle.Annotations" Version="0.28.0" />
+		<PackageReference Update="D2L.Services.Core.Exceptions" Version="2.0.1.15" />
+		<PackageReference Update="FluentAssertions" Version="6.2.0" />
 		<PackageReference Update="Microsoft.AspNet.WebApi.Core" Version="[5.2.7]" />
 		<PackageReference Update="Microsoft.AspNet.WebApi.OwinSelfHost" Version="[5.2.7]" />
```
j3parker, line 10: Keeping the old-school WebAPI references pinned because of framework constraints.
(This package doesn't exist at all in .NET Core)

Packages.props
```diff
@@ -1,23 +1,23 @@
...
-		<PackageReference Update="Microsoft.Extensions.Caching.Memory" Version="[5.0.0]" />
-		<PackageReference Update="Microsoft.NET.Test.Sdk" Version="[17.0.0]" />
-		<PackageReference Update="Moq" Version="[4.16.1]" />
+		<PackageReference Update="Microsoft.Extensions.Caching.Memory" Version="5.0.0" />
+		<PackageReference Update="Microsoft.NET.Test.Sdk" Version="17.0.0" />
+		<PackageReference Update="Moq" Version="4.16.1" />
 		<PackageReference Update="Newtonsoft.Json" Version="[6.0.8]" />
```
j3parker, line 14: Keeping Newtonsoft pinned because of insanity.

I tried a few things moving this to Directory.Build.props in #201 but that stuff I think is checking version compatability better, lol -- so a bunch of our test support packages are angry at the very old Newtonsoft. Does that make sense @mthjones ?

But the tests happen to pass such as they are, so I'm not super worried... but it is a mess.

j3parker:
Mostly annoyed at these: 
image

Packages.props
omsmith: Hmm, didn't realize that's what the brackets meant.

omsmith (approved):

Packages.props
mthjones: > Keeping Newtonsoft pinned because of insanity.

Could you expand on what the "insanity" is? There generally shouldn't be a need to pin Newtonsoft.Json here unless you know that future releases don't work. The responsibility should typically fall on the consumer of the package to manage binding redirects/dependency versions as appropriate, rather than raising warnings/errors during restore:

```
C:\Users\mjones\code\dotnet-scratch\oauth2\testproj\testproj.csproj : warning NU1608: Detected package version outside of dependency constraint: D2L.Security.OAuth2 11.9.1.462 requires Newtonsoft.Json (= 6.0.8) but version Newtonsoft.Json 13.0.1 was resolved.
```

But as to the change in #201...I don't see `<ManagePackageVersionsCentrally>true</ManagePackageVersionsCentrally>` being set and it was still using `PackageReference Include="..."` in the `Directory.Packages.props`, so it was essentially including every package in every project. Not sure if that's what you intended?

If you were looking to switch to NuGet's central package versions, you have to opt-in with that property (probably in `Directory.Build.props`) and switch those items in `Directory.Packages.props` to `<PackageVersion Include="..."` instead. That might have been what it was upset about.

j3parker: RE: Newtonsoft, 7 and forward include changes to deserialization behaviour. This library deserializes JSON from 3rd parties (like, our customers) as an internal impl detail. I'm not sure off-hand what the implications of all those changes are to this library/we don't test for it. Additionally we want to get off Newtonsoft in favour of System.Text.Json (we've done that for the .NET 5 build for now).

Ah -- `ManagePackageVersionsCentrally` -- that explains things. The other project I was copying stuff from does that in `Directory.Build.props` and I missed that. Thanks!

PR #221, opened 2022-02-10
Updating D2L.CodeStyle.Analyzers to v0.186.0

JeffAshton:


src/D2L.Security.OAuth2/Provisioning/Default/AuthServiceClient.cs
```diff
@@ -8,16 +9,21 @@
 using D2L.Services.Core.Exceptions;
 
 #if NET5_0
-using JsonPropertyName = System.Text.Json.Serialization.JsonPropertyNameAttribute;
+using JsonPropertyNameAttribute = System.Text.Json.Serialization.JsonPropertyNameAttribute;
```
JeffAshton, line 12: Updated to match actual name.

omsmith (approved):

PR #233, opened 2022-04-06
Remove some .NET Framework references

cesar-d2l:
This removes `System.IdentityModel` which wasn't being referenced anywhere

The motivation for doing this: We want to remove `System.IdentityModel` from the LMS (well all `References`) in our transition to .NET. However I have been noticing certain projects building that use classes in `System.IdentityModel` but don't reference `System.IdentityModel`. We think it's because it's picking up `System.IdentityModel` as a transitive dependency. For example, [D2L.LP.Auth](https://github.com/Brightspace/lms/blob/master/platformTools/auth/D2L.LP.Auth/D2L.LP.Auth.csproj#L24) depends on this package as well as `System.IdentityModel`. Even though I remove `System.IdentityModel` reference, the project still compiles :(

### DIFF
--- a/src/D2L.Security.OAuth2/D2L.Security.OAuth2.csproj
+++ b/src/D2L.Security.OAuth2/D2L.Security.OAuth2.csproj
@@ -19,6 +19,7 @@
 		<PackageReference Include="D2L.Services.Core.Exceptions" />
 		<PackageReference Include="System.Collections.Immutable" />
 		<PackageReference Include="System.IdentityModel.Tokens.Jwt" />
+		<PackageReference Include="System.Net.Http" />
 	</ItemGroup>
 
 	<ItemGroup Condition="'$(TargetFramework)' == 'net50'">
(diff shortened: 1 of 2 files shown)


src/D2L.Security.OAuth2/D2L.Security.OAuth2.csproj
```diff
@@ -19,6 +19,7 @@
 		<PackageReference Include="D2L.Services.Core.Exceptions" />
 		<PackageReference Include="System.Collections.Immutable" />
 		<PackageReference Include="System.IdentityModel.Tokens.Jwt" />
+		<PackageReference Include="System.Net.Http" />
```
j3parker, line 22: This is an unmaintained package (last published 2018), its part of the standard library now.

This library multi-targets .NET 4.7.2 and .NET 5 and runs in Linux on .NET 6 just fine.

j3parker (approved):

cesar-d2l:
Can anyone with write access merge this in?

j3parker:
Thanks!

PR #239, opened 2022-05-06
Add acceptable aud claim value as constant

sammeyerreed:
New constant to be used in https://github.com/Brightspace/oauth2/pull/1245


j3parker:
I'm fine with this going here but it mostly is a thing for the auth service, like clients don't need to know about it (this is the client library)

It's true that clients could use it, but if they're using this library they can and will use the standard one.

omsmith:
I would prefer to keep this to a constant in the service.

sammeyerreed:
Ok I'll close this, happy to keep it in the service repo

PR #245, opened 2022-05-18
Reject RSA JSON key whose modulus has 0 byte prefix

jcousins-cs:
As per https://datatracker.ietf.org/doc/html/rfc7518#section-6.3.1.1, we should reject RSA JSON web keys with decoded modulus that starts with a 0 byte.

### DIFF
--- a/src/D2L.Security.OAuth2/Keys/JsonWebKey.cs
+++ b/src/D2L.Security.OAuth2/Keys/JsonWebKey.cs
@@ -104,12 +104,17 @@ namespace D2L.Security.OAuth2.Keys {
 						throw new JsonWebKeyParseException( "RSA JSON web key has private key material" );
 					}
 
-					return new RsaJsonWebKey(
+					if( !RsaJsonWebKey.TryParse(
 						id: id,
 						expiresAt: expiresAt,
-						n: data[ "n" ].ToString(),
-						e: data[ "e" ].ToString()
-					);
+						n: data["n"].ToString(),
+						e: data["e"].ToString(),
+						rsaJsonWebKey: out RsaJsonWebKey rsaJsonWebKey
+					) ) {
+						throw new JsonWebKeyParseException( "decoded 'n' parameter starts with 0 byte in RSA JSON web key" );
+					}
+
+					return rsaJsonWebKey;
 
 				case "EC":
 					if( !data.ContainsKey( "crv" ) ) {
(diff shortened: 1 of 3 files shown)


src/D2L.Security.OAuth2/Keys/JsonWebKey.cs
```diff
@@ -104,12 +104,17 @@ public abstract class JsonWebKey {
...
 						id: id,
 						expiresAt: expiresAt,
-						n: data[ "n" ].ToString(),
-						e: data[ "e" ].ToString()
-					);
+						n: data["n"].ToString(),
+						e: data["e"].ToString(),
+						rsaJsonWebKey: out RsaJsonWebKey rsaJsonWebKey
+					) ) {
+						throw new JsonWebKeyParseException( "decoded 'n' parameter starts with 0 byte in RSA JSON web key" );
```
j3parker, line 114: nit: it would probably be better for `RsaJsonWebKey` to simply throw, all things considered. the `TryX` pattern kinda has the flaw of being opaque about what caused a thing to fail, so when you do want details (e.g. to format a specific error) you need another `out` to explain what the failure was.

That being said I'm fine with this for now. We should refactor this library to not use exceptions eventually, and I could grab this one later.

test/D2L.Security.OAuth2.UnitTests/Keys/JsonWebKeyTests.cs
```diff
@@ -61,6 +61,15 @@ internal sealed class JsonWebKeyTests {
...
 		}
 
+		[Test]
+		public void FromJson_Modulus0PrefixThrows() {
+			string id = Guid.NewGuid().ToString();
+			string example =
+				@"{""kid"":""" + id + @""",""kty"":""RSA"",""use"":""sig"",""n"":""AKYl5hffy9FDuCuQD8x6ojmJfyrVVwID5VR4VKKT87WlvyUYdpPTA3rqSANXABfUv6-akQVnANUGj9J5RmpcN45v30h77VGGIcFIRfV-iNfVUZr_wV_TXtm7nbeXWnyPQ2zbNi7atF_8nzNXVC_Nb2kdK8-42GkOWxdgkUB1QnCIVpIpPOTxuXNciv_wLPONlqqIAaVvPs7BKAerOdxTaQxt-_x7GQAJ4AE5k2NQLeDu4wuqfnNClTw39uaC1TDlkLm8lYAhPRZbFcaax1qrF3p5ZxxJoWXxGxQf-SgeoY9sle-44by0bJTb8in5RQbOgBgonMzQ85qFx9hldxQFSxU"",""e"":""AQAB""}";
+
+			Assert.Throws<JsonWebKeyParseException>( () => JsonWebKey.FromJson( example ) );
+		}
```
j3parker, line 71: Thanks

j3parker (approved):

PR #257, opened 2022-08-02
DE49363 - Support additional data in jwks

maggiesazio:
https://rally1.rallydev.com/#/?detail=/defect/643330303623&fdp=true

### DIFF
--- a/src/D2L.Security.OAuth2/Keys/Default/JsonWebKeySet.cs
+++ b/src/D2L.Security.OAuth2/Keys/Default/JsonWebKeySet.cs
@@ -1,6 +1,7 @@
 ﻿using System;
 using System.Collections.Generic;
 using System.Collections.Immutable;
+using System.Linq;
 using D2L.Security.OAuth2.Utilities;
 
 #if NET6_0
@@ -23,9 +24,17 @@ namespace D2L.Security.OAuth2.Keys.Default {
 					throw new JsonWebKeyParseException( "invalid json web key set: missing keys array" );
 				}
 
+				#if NET6_0
+				System.Text.Json.JsonElement keysElement = ( System.Text.Json.JsonElement )data["keys"];
+				if ( keysElement.ValueKind is not System.Text.Json.JsonValueKind.Array ) {
+					throw new JsonWebKeyParseException("invalid json web key set: keys not an array");
+				}
+				List<System.Text.Json.JsonElement> keyObjects = keysElement.EnumerateArray().ToList();
+				#else
 				if ( data["keys"] is not IEnumerable<object> keyObjects ) {
 					throw new JsonWebKeyParseException( "invalid json web key set: keys not an array" );
 				}
+				#endif
 
 				var builder = ImmutableArray.CreateBuilder<JsonWebKey>();
 				foreach( object keyObject in keyObjects ) {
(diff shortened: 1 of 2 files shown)


omsmith (approved):

src/D2L.Security.OAuth2/Keys/Default/JsonWebKeySet.cs
```diff
@@ -17,13 +17,15 @@ internal sealed class JsonWebKeySet {
...
-				var data = JsonSerializer.Deserialize<Dictionary<string, List<object>>>( json );
+				var data = JsonSerializer.Deserialize<Dictionary<string, object>>( json );
 
 				if( !data.ContainsKey( "keys" ) ) {
 					throw new JsonWebKeyParseException( "invalid json web key set: missing keys array" );
 				}
 
-				List<object> keyObjects = data["keys"];
+				if ( data["keys"] is not IEnumerable<object> keyObjects ) {
+					throw new JsonWebKeyParseException( "invalid json web key set: keys not an array" );
```
maggiesazio, line 27: I have no idea why tests are failing CI because of this. They're passing locally. Any ideas what I should try?

maggiesazio: Thanks. Looks like a .net 6 issue where it's a JsonElement.

src/D2L.Security.OAuth2/Keys/Default/JsonWebKeySet.cs
```diff
@@ -17,13 +18,23 @@ internal sealed class JsonWebKeySet {
...
 					throw new JsonWebKeyParseException( "invalid json web key set: missing keys array" );
 				}
 
-				List<object> keyObjects = data["keys"];
+				#if NET6_0
+				System.Text.Json.JsonElement keysElement = ( System.Text.Json.JsonElement )data["keys"];
+				if ( keysElement.ValueKind is not System.Text.Json.JsonValueKind.Array ) {
+					throw new JsonWebKeyParseException("invalid json web key set: keys not an array");
+				}
+				List<System.Text.Json.JsonElement> keyObjects = keysElement.EnumerateArray().ToList();
```
j3parker, line 32: For .NET 6 we might want to just switch to `JsonDocument` in .NET 6 vs. `JsonSerializer.Deserialize`+the casts/reflection, but I could take that for later, this is fine.

j3parker (approved):

maggiesazio:
How does the LMS get this change? We wanted to get the defect fixed for 20.22.8 but I have no idea how to that with this library.

j3parker:
> How does the LMS get this change?

Typically via dependabot, but you can [do it manually if you want](https://github.com/Brightspace/lms/blob/c6afbebbb577620aaf48706f41b41a5a7124e53b/Directory.Packages.props#L42-L43).

maggiesazio:
> > How does the LMS get this change?
> 
> Typically via dependabot, but you can [do it manually if you want](https://github.com/Brightspace/lms/blob/c6afbebbb577620aaf48706f41b41a5a7124e53b/Directory.Packages.props#L42-L43).

Thanks. Looks like the LMS is a little behind so I'll probably do it manually. I figure it will have to be manual to get it in to 20.22.8 anyways. Can you merge the PR for me please?

PR #292, opened 2023-02-24
dispose of keys when they fall out of cache

omsmith:


j3parker (approved):

src/D2L.Security.OAuth2/Keys/Caching/InMemoryPublicKeyCache.net4x.cs
```diff
@@ -26,5 +27,13 @@ internal sealed partial class InMemoryPublicKeyCache : IInMemoryPublicKeyCache {
...
 			return result;
 		}
+
+		private static void EvictionCallback( CacheEntryRemovedArguments args ) {
+			if( args.CacheItem.Value is not D2LSecurityToken key ) {
+				return;
+			}
+
+			key.Dispose();
+		}
```
omsmith, line 37: Can check EvictionReason here and in the net6 path to not dispose on explicit removals or something, but I figure if we do the `Ref()` within the `Set` methods here, then the cache owns the ref and should do the `Dispose()` whenever it leaves.

src/D2L.Security.OAuth2/Validation/AccessTokens/AccessTokenValidator.cs
```diff
@@ -61,9 +61,10 @@ string token
 
 			string keyId = unvalidatedToken.Header[ "kid" ].ToString();
 
-			D2LSecurityToken signingKey = await m_publicKeyProvider
+			using D2LSecurityToken signingKey = ( await m_publicKeyProvider
 				.GetByIdAsync( keyId )
-				.ConfigureAwait( false );
+				.ConfigureAwait( false )
+			).Ref();
```
omsmith, line 67: Need to `Ref()` and dispose here now so the cache disposing doesn't fully dispose it during usage.

PR #295, opened 2023-03-03
consistently avoid double-slash when joining paths

omsmith:


src/D2L.Security.OAuth2/Keys/Default/Data/JwksProvider.cs
```diff
@@ -86,20 +86,7 @@ Uri jwkEndpoint
...
-
-		private static string MakeSureThereIsATrailingSlash( Uri uri ) {
-			string root = uri.ToString();
-			if( root[root.Length - 1] == '/' ) {
-				return root;
-			}
-
-			return root + '/';
+			return authEndpoint.RelativePathAsNonLeaf( HttpUtility.UrlEncode( keyId ) );
 		}
```
omsmith, line 90: This was already fine. Just sharing the code

j3parker (approved):

src/D2L.Security.OAuth2/Provisioning/Default/AuthServiceClient.cs
```diff
@@ -60,7 +60,7 @@ Uri authEndpoint
 			}
 
 			m_client = httpClient;
-			m_tokenProvisioningEndpoint = new Uri( authEndpoint + TOKEN_PATH );
+			m_tokenProvisioningEndpoint = authEndpoint.RelativePathAsNonLeaf( TOKEN_PATH );
```
omsmith, line 63: This would result in `/foo//connect/token` depending on the input authEndpoint.

This was addressed before in-product by removing the trailing slash from the input, but looking to handle it regardless here.

PR #312, opened 2023-07-07
Changes Needed to Generate Sync RequestAuthenticator

AnaCoda:
[US154650](https://rally1.rallydev.com/#/?detail=/userstory/706189919605&fdp=true): Async to Sync Generator: Generate RequestAuthenticator
Added [GenerateSync] as needed, partial classes, and small changes (made Task<ID2LPrincipal> IRequestAuthenticator.AuthenticateAsync() asynchronous)

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -22,7 +22,7 @@
 	</PropertyGroup>
 
 	<PropertyGroup>
-		<LangVersion>9.0</LangVersion>
+		<LangVersion>11.0</LangVersion>
 		<ManagePackageVersionsCentrally>true</ManagePackageVersionsCentrally>
 	</PropertyGroup>
 </Project>
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -24,5 +24,6 @@
 	<PropertyGroup>
 		<LangVersion>11.0</LangVersion>
 		<ManagePackageVersionsCentrally>true</ManagePackageVersionsCentrally>
+		<WarningsAsErrors>CS8785</WarningsAsErrors>
 	</PropertyGroup>
 </Project>
--- a/src/D2L.Security.OAuth2.TestFramework/TestAccessToken.cs
+++ b/src/D2L.Security.OAuth2.TestFramework/TestAccessToken.cs
@@ -32,7 +32,7 @@ namespace D2L.Security.OAuth2.TestFramework {
 
 			claims.Add( new Claim( Constants.Claims.ISSUER, DEFAULT_ISSUER ) );
 
-			using( var httpClient = new D2LHttpClient() ) {
+			using( var httpClient = new HttpClient() ) {
 				IAccessTokenProvider provider = TestAccessTokenProviderFactory.Create( httpClient, tokenProvisioningEndpoint );
 				IAccessToken token = await provider.ProvisionAccessTokenAsync( claims, scopes ).ConfigureAwait( false );
 				return token.Token;
--- a/src/D2L.Security.OAuth2.TestFramework/TestAccessToken.cs
+++ b/src/D2L.Security.OAuth2.TestFramework/TestAccessToken.cs
@@ -6,8 +6,6 @@ using System.Security.Claims;
 using System.Threading.Tasks;
 using D2L.Security.OAuth2.Provisioning;
 using D2L.Security.OAuth2.Scopes;
-using D2L.Security.OAuth2.Utilities;
-using D2L.Services;
 using IAccessToken = D2L.Security.OAuth2.Provisioning.IAccessToken;
 
 namespace D2L.Security.OAuth2.TestFramework {
--- a/src/D2L.Security.OAuth2.TestFramework/TestAccessToken.cs
+++ b/src/D2L.Security.OAuth2.TestFramework/TestAccessToken.cs
@@ -6,6 +6,7 @@ using System.Security.Claims;
 using System.Threading.Tasks;
 using D2L.Security.OAuth2.Provisioning;
 using D2L.Security.OAuth2.Scopes;
+using D2L.Security.OAuth2.Utilities;
 using D2L.Services;
 using IAccessToken = D2L.Security.OAuth2.Provisioning.IAccessToken;
 
@@ -31,7 +32,7 @@ namespace D2L.Security.OAuth2.TestFramework {
 
 			claims.Add( new Claim( Constants.Claims.ISSUER, DEFAULT_ISSUER ) );
 
-			using( var httpClient = new HttpClient() ) {
+			using( var httpClient = new D2LHttpClient() ) {
 				IAccessTokenProvider provider = TestAccessTokenProviderFactory.Create( httpClient, tokenProvisioningEndpoint );
 				IAccessToken token = await provider.ProvisionAccessTokenAsync( claims, scopes ).ConfigureAwait( false );
 				return token.Token;
--- a/src/D2L.Security.OAuth2/Keys/Default/Data/IJwksProvider.cs
+++ b/src/D2L.Security.OAuth2/Keys/Default/Data/IJwksProvider.cs
@@ -1,5 +1,5 @@
-﻿using D2L.CodeStyle.Annotations;
-using System.Threading.Tasks;
+﻿using System.Threading.Tasks;
+using D2L.CodeStyle.Annotations;
 
 namespace D2L.Security.OAuth2.Keys.Default.Data {
 	internal partial interface IJwksProvider {
--- a/src/D2L.Security.OAuth2/Keys/Default/Data/IJwksProvider.cs
+++ b/src/D2L.Security.OAuth2/Keys/Default/Data/IJwksProvider.cs
@@ -1,8 +1,11 @@
-﻿using System.Threading.Tasks;
+﻿using D2L.CodeStyle.Annotations;
+using System.Threading.Tasks;
 
 namespace D2L.Security.OAuth2.Keys.Default.Data {
-	internal interface IJwksProvider {
+	internal partial interface IJwksProvider {
+		[GenerateSync]
 		Task<JsonWebKeySet> RequestJwksAsync();
+		[GenerateSync]
 		Task<JsonWebKeySet> RequestJwkAsync( string keyId );
 		string Namespace { get; }
 	}
--- a/src/D2L.Security.OAuth2/Keys/Default/Data/JwksProvider.cs
+++ b/src/D2L.Security.OAuth2/Keys/Default/Data/JwksProvider.cs
@@ -30,7 +30,7 @@ namespace D2L.Security.OAuth2.Keys.Default.Data {
 			try {
 				using( HttpResponseMessage response = await m_httpClient.GetAsync( m_jwksEndpoint ).ConfigureAwait( false ) ) {
 					response.EnsureSuccessStatusCode();
-					string jsonResponse = await response.Content.ReadAsStringAsync().ConfigureAwait(false);
+					string jsonResponse = await response.Content.ReadAsStringAsync().ConfigureAwait( false );
 					var jwks = new JsonWebKeySet( jsonResponse, m_jwksEndpoint );
 					return jwks;
 				}
--- a/src/D2L.Security.OAuth2/Keys/Default/Data/JwksProvider.cs
+++ b/src/D2L.Security.OAuth2/Keys/Default/Data/JwksProvider.cs
@@ -1,19 +1,22 @@
 ﻿using System;
+using System.IO;
 using System.Net;
 using System.Net.Http;
 using System.Threading.Tasks;
 using System.Web;
+using D2L.CodeStyle.Annotations;
+using D2L.Security.OAuth2.Utilities;
 using D2L.Security.OAuth2.Validation.Exceptions;
 using D2L.Services;
 
 namespace D2L.Security.OAuth2.Keys.Default.Data {
-	internal sealed class JwksProvider : IJwksProvider {
-		private readonly HttpClient m_httpClient;
+	internal sealed partial class JwksProvider : IJwksProvider {
+		private readonly D2LHttpClient m_httpClient;
 		private readonly Uri m_jwksEndpoint;
 		private readonly Uri m_jwkEndpoint;
 
 		public JwksProvider(
-			HttpClient httpClient,
+			D2LHttpClient httpClient,
 			Uri jwksEndpoint,
 			Uri jwkEndpoint
 		) {
@@ -22,11 +25,15 @@ namespace D2L.Security.OAuth2.Keys.Default.Data {
 			m_jwkEndpoint = jwkEndpoint;
 		}
 
+		[GenerateSync]
 		async Task<JsonWebKeySet> IJwksProvider.RequestJwksAsync() {
 			try {
 				using( HttpResponseMessage response = await m_httpClient.GetAsync( m_jwksEndpoint ).ConfigureAwait( false ) ) {
 					response.EnsureSuccessStatusCode();
-					string jsonResponse = await response.Content.ReadAsStringAsync().ConfigureAwait( false );
+					var reader = new StreamReader(await response.Content.ReadAsStreamAsync().ConfigureAwait(false));
+					string jsonResponse = await reader
+						.ReadToEndAsync()
+						.ConfigureAwait(false);
 					var jwks = new JsonWebKeySet( jsonResponse, m_jwksEndpoint );
 					return jwks;
 				}
@@ -37,6 +44,7 @@ namespace D2L.Security.OAuth2.Keys.Default.Data {
 			}
 		}
 
+		[GenerateSync]
 		async Task<JsonWebKeySet> IJwksProvider.RequestJwkAsync( string keyId ) {
 			var url = GetJwkEndpoint( m_jwkEndpoint, keyId );
 			if( url == null ) {
@@ -61,9 +69,10 @@ namespace D2L.Security.OAuth2.Keys.Default.Data {
 
 					res.EnsureSuccessStatusCode();
 
-					string json = await res.Content
-						.ReadAsStringAsync()
-						.ConfigureAwait( false );
+					var reader = new StreamReader(await res.Content.ReadAsStreamAsync().ConfigureAwait(false));
+					string json = await reader
+						.ReadToEndAsync()
+						.ConfigureAwait(false);
 
 					JsonWebKey jwk = JsonWebKey.FromJson( json );
 					return new JsonWebKeySet( jwk, url );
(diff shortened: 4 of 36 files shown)


src/D2L.Security.OAuth2/Keys/Default/Data/JwksProvider.cs
```diff
@@ -22,11 +25,15 @@ Uri jwkEndpoint
 			m_jwkEndpoint = jwkEndpoint;
 		}
 
+		[GenerateSync]
 		async Task<JsonWebKeySet> IJwksProvider.RequestJwksAsync() {
 			try {
 				using( HttpResponseMessage response = await m_httpClient.GetAsync( m_jwksEndpoint ).ConfigureAwait( false ) ) {
 					response.EnsureSuccessStatusCode();
-					string jsonResponse = await response.Content.ReadAsStringAsync().ConfigureAwait( false );
+					var reader = new StreamReader(await response.Content.ReadAsStreamAsync().ConfigureAwait(false));
```
j3parker, line 33:
```suggestion
					using var reader = new StreamReader(await response.Content.ReadAsStreamAsync().ConfigureAwait(false));
```

(ditto elsewhere)

src/D2L.Security.OAuth2/Keys/Development/InMemoryPublicKeyDataProvider.cs
```diff
@@ -14,30 +15,34 @@ namespace D2L.Security.OAuth2.Keys.Development {
...
-			return Task.CompletedTask;
+			await Task.CompletedTask.ConfigureAwait(false);
 		}
 
-		Task IPublicKeyDataProvider.DeleteAsync( Guid id ) {
+		[GenerateSync]
+		async Task IPublicKeyDataProvider.DeleteAsync( Guid id ) {
 			m_keys.TryRemove( id, out JsonWebKey removedKey );
-			return Task.CompletedTask;
+			await Task.CompletedTask.ConfigureAwait(false);
```
j3parker, line 45: Hm, we could transform `return Task.CompletedTask;` into just `return;` I think.

[Check out the decompilation of](https://sharplab.io/#v2:EYLgtghglgdgPgAQAwAIEEYCsBuAsAKAIQGY10A2NAJhQGEUBvAlFtUjShADjUoDEA9gIAUASkbNWUhAE5eAOloCwABwA2AUwAuGgCYJyigTABmUAOYBXAE4aAggHdoW4ShMQ1AZw0pRefFIAvgSBQA=) a method with `await Task.CompletedTask.ConfigureAwait( false );`. It's ultimately not really doing much else but it's definitely more junk than just `return Task.CompletedTask`. Mostly though it's just "weird" to see `await Task.CompletedTask` :) I'd like for `[GenerateSync]` to allow us to write idiomatic code.

src/D2L.Security.OAuth2/Utilities/D2LHttpClient.cs
```diff
@@ -0,0 +1,125 @@
...
+			=> m_httpClient.SendAsync(request, completionOption);
+
+		public Task<HttpResponseMessage> SendAsync(HttpRequestMessage request, HttpCompletionOption completionOption, CancellationToken cancellationToken)
+			=> m_httpClient.SendAsync(request, completionOption, cancellationToken);
+
+#pragma warning disable D2L0018 // Avoid using dangerous methods
+		public HttpResponseMessage Get(string requestUri, HttpCompletionOption completionOption, CancellationToken cancellationToken) {
+			var task = Task.Run(() => m_httpClient.GetAsync(requestUri, completionOption, cancellationToken));
+			task.Wait();
+			return task.Result;
```
j3parker, line 60: I'd be fine with just

```cs
=> Task.Run( () => m_httpClient.getAsync( requestUri, completionOption, cancellationToken ) ).Result;
```

but no big deal

src/D2L.Security.OAuth2/Validation/AccessTokens/AccessTokenValidatorFactory.cs
```diff
@@ -37,7 +38,7 @@ IPublicKeyDataProvider publicKeyDataProvider
 		/// <param name="jwkEndpoint">The full URI of the remote JWK path</param>
 		/// <returns>A new <see cref="IAccessTokenValidator"/></returns>
 		public static IAccessTokenValidator CreateRemoteValidator(
-			HttpClient httpClient,
+			D2LHttpClient httpClient,
```
j3parker, line 41: I think the public methods should take `ID2LHttpClient`?

src/D2L.Security.OAuth2/Keys/Development/InMemoryPublicKeyDataProvider.cs
j3parker: I wouldn't add the `async` modifier to them, e.g. I'd hope that

```cs
Task IPublicKeyDataProvider.DeleteAsync( Guid id ) {
	m_keys.TryRemove( id, out JsonWebKey removedKey );
	return Task.CompletedTask;
}
```

could get transformed into

```cs
void IPublicKeyDataProvider.Delete( Guid id ) {
	m_keys.TryRemove( id, out JsonWebKey removedKey );
	return;
}
```

or something like that

src/D2L.Security.OAuth2/Keys/Default/Data/JwksProvider.cs
```diff
@@ -22,11 +25,12 @@ Uri jwkEndpoint
 			m_jwkEndpoint = jwkEndpoint;
 		}
 
+		[GenerateSync]
 		async Task<JsonWebKeySet> IJwksProvider.RequestJwksAsync() {
 			try {
 				using( HttpResponseMessage response = await m_httpClient.GetAsync( m_jwksEndpoint ).ConfigureAwait( false ) ) {
 					response.EnsureSuccessStatusCode();
-					string jsonResponse = await response.Content.ReadAsStringAsync().ConfigureAwait( false );
+					string jsonResponse = await response.Content.ReadAsStringAsync().ConfigureAwait(false);
```
j3parker, line 33:
```suggestion
					string jsonResponse = await response.Content.ReadAsStringAsync().ConfigureAwait( false );
```

(nit, mostly to undo the change)

j3parker (approved):

src/D2L.Security.OAuth2/Utilities/D2LHttpClient.cs
```diff
@@ -0,0 +1,125 @@
+﻿using System;
+using System.Net.Http;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace D2L.Security.OAuth2.Utilities
+{
+	public class D2LHttpClient : ID2LHttpClient
```
omsmith, line 8: Didn't our strategy pivot to sticking with `HttpClient` and generating the `Task.Run` wrapper?

AnaCoda: Correct me if I'm wrong @j3parker but I think this is a form of future-proofing so that we can implement true "sync" versions of HttpClient methods in the future. We do have the ability to wrap them in Task.Run but saved that for one-offs and not the HttpClient class

j3parker: I think the interface we're going to need to wrap HTTP is going to look a lot different than what we've got here (like methods like `T GetAsJson<T>( string url );` etc., high level stuff).

I'm OK if we can drop D2LHttpClient for now and rely on the generator, or if that's a hassle I don't care if we have this (just it'll change drastically in a bit, that's all.)

AnaCoda: It is a bit of a hassle, because given the common names "Get" and "Send", I would have to check the types of whatever is calling them. Anyway, it would have to be a new package so if things are going to change drastically anyway I feel that we shouldn't make additional changes to the generator only to revert them later

AnaCoda:
> Very cool.
> 
> Could we try `[GenerateSync]` on one of the tests for fun? If it works to generate a sync test that would be pretty cool, heh. If it doesn't that's fine for now but I'd be curious about why.

Any tests for these changes will be in https://github.com/Brightspace/D2L.Security.OAuth2/pull/316/ (force pushed to remove them from this branch)

src/D2L.Security.OAuth2/Utilities/D2LHttpClient.cs
omsmith: I guess I'm unsure why we added the `ReadAsStringAsync` wrapping into the generator, when it was directly related to this HttpClient usage - and I thought we were ultimately adding it to avoid having to wrap HttpClient (at the moment, at least).

Whatever works if we're just working incrementally though. I'd rather us not cut a release of D2L.Security.OAuth2 as-is though.

omsmith, line 8:
```suggestion
	public sealed class D2LHttpClient : ID2LHttpClient
```

omsmith, line 8:
```suggestion
	public sealed class D2LHttpClient : ID2LHttpClient
```

src/D2L.Security.OAuth2/Utilities/D2LHttpClient.cs
```diff
@@ -0,0 +1,125 @@
...
+
+namespace D2L.Security.OAuth2.Utilities
+{
+	public class D2LHttpClient : ID2LHttpClient
+	{
+		/// <summary>
+		/// This class implements sync versions of HTTPClient methods.
+		/// TODO: Replace the sync methods which are currently using Task.Run(async verison); task.Wait();
+		/// With (probably) HttpWebRequest
+		/// </summary>
```
omsmith, line 14: I think this comment belongs more on the `D2LHttpClient` class, rather than `m_httpClient`?

src/D2L.Security.OAuth2/Utilities/D2LHttpClient.cs
AnaCoda: I see, it's because ReadAsStringAsync is part of HttpContent not HttpClient which complicates things

src/D2L.Security.OAuth2/Utilities/D2LHttpClient.cs
```diff
@@ -0,0 +1,125 @@
...
+
+		public D2LHttpClient() {
+			m_httpClient = new HttpClient();
+		}
+		public D2LHttpClient(HttpMessageHandler handler) {
+			m_httpClient = new HttpClient(handler);
+		}
+		public D2LHttpClient(HttpMessageHandler handler, bool disposeHandler) { 
+			m_httpClient = new HttpClient(handler, disposeHandler);
+		}
```
omsmith, line 25: For the moment I would suggest making `ID2LHttpClient` and `D2LHttpClient` `internal`, using it to wrap the passed in `HttpClient`, and avoid changing the public interface of this library (as far as the factories go and whatnot (i.e. undo changes to `AccessTokenProviderFactory`)).

```suggestion
		public D2LHttpClient( HttpClient httpClient ) {
			m_httpClient = httpClient;
		}
```

src/D2L.Security.OAuth2/Utilities/D2LHttpClient.cs
```diff
@@ -0,0 +1,125 @@
...
+		public Task<HttpResponseMessage> SendAsync(HttpRequestMessage request, CancellationToken cancellationToken)
+			=> m_httpClient.SendAsync(request, cancellationToken);
+
+		public Task<HttpResponseMessage> SendAsync(HttpRequestMessage request, HttpCompletionOption completionOption)
+			=> m_httpClient.SendAsync(request, completionOption);
+
+		public Task<HttpResponseMessage> SendAsync(HttpRequestMessage request, HttpCompletionOption completionOption, CancellationToken cancellationToken)
+			=> m_httpClient.SendAsync(request, completionOption, cancellationToken);
+
+#pragma warning disable D2L0018 // Avoid using dangerous methods
```
omsmith, line 56: Would be great if we could use `[DangerousMethods.Audited( ... )] instead of disabling the `#pragma`

src/D2L.Security.OAuth2/Utilities/ID2LHttpClient.cs
```diff
@@ -0,0 +1,48 @@
...
+using System.Threading.Tasks;
+
+namespace D2L.Security.OAuth2.Utilities
+{
+	public interface ID2LHttpClient : IDisposable
+	{
+		/// <summary>
+		/// This interface was created to allow the sync generator in D2L.CodeStyle to strip the async suffix from HttpClient calls
+		/// when it tries to convert async code to sync code.
+		/// </summary>
```
omsmith, line 13: Comment belongs on `ID2LHttpClient`, not the method.

src/D2L.Security.OAuth2/Validation/Request/RequestAuthenticator.cs
```diff
@@ -14,14 +15,16 @@ internal sealed partial class RequestAuthenticator : IRequestAuthenticator {
...
 
-		Task<ID2LPrincipal> IRequestAuthenticator.AuthenticateAsync(
+		[GenerateSync]
+		async Task<ID2LPrincipal> IRequestAuthenticator.AuthenticateAsync(
 			HttpRequestMessage request
 		) {
 			string bearerToken = request.GetBearerTokenValue();
 
-			return AuthenticateAsync( bearerToken );
+			return await AuthenticateAsync( bearerToken ).ConfigureAwait(false);
```
omsmith, line 24: I think the latest version of the generator doesn't require this change to `async`, right?

Directory.Build.props
```diff
@@ -24,5 +24,6 @@
 	<PropertyGroup>
 		<LangVersion>11.0</LangVersion>
 		<ManagePackageVersionsCentrally>true</ManagePackageVersionsCentrally>
+		<WarningsAsErrors>CS8785</WarningsAsErrors>
```
omsmith, line 27: What is CS8785?

omsmith (commented):
There's a number of cases of new code not matching our existing style. Could you get that fixed up?

* opening braces shouldn't have their own line (`class Foo : IFoo {`)
* method declarations and invocations should have spaces within the parens (`void Foo( string bar ) {`, `Foo( "Bar" );`)

Directory.Build.props
AnaCoda: Generator failed to generate source (want to make sure build fails if the sync generator fails). Should I comment in here?

omsmith: No that's fine, just new to me and wanted to be sure it made sense. I think we'd also notice it if we manually defined the interface methods instead of using `[GenerateSync]` to also fill those out.

```csharp
public interface IRequestAuthenticator {
  Task<ID2LPrincipal> AuthenticateAsync( HttpRequestMessage request );
  ID2LPrincipal Authenticate( HttpRequestMessage request );
}

internal sealed partial class RequestAuthenticator : IRequestAuthenticator {
  [GenerateSync]
  Task<ID2LPrincipal> IRequestAuthenticator.AuthenticateAsync(
    HttpRequestMessage request
  ) { ... }
```

In the event of a failure it would complain that `RequestAuthenticator` doesn't implement the interface. Don't know if @j3parker has a preference.

j3parker: Generating the sync interface for you certainly makes it easier to maintain... I dno, I don't really understand why that diagnostic isn't already an error. Like why would you be running a generator if you were cool with it crashing? lol

src/D2L.Security.OAuth2/Utilities/D2LHttpClient.cs
AnaCoda: Is this an existing decorator or are you proposing its creation

Directory.Build.props
j3parker: We also insert `[Blocking]` on the generated stuff, etc. FYI

Like if we were having you specify them manually I think we'd also want to analyze them to make sure they matched what we wanted... and at that point it seems simpler to just generate.

src/D2L.Security.OAuth2/Utilities/D2LHttpClient.cs
omsmith: Existing, just typo'd it:

https://search.d2l.dev/search?project=lms&full=DangerousMethodUsage

src/D2L.Security.OAuth2.TestFramework/TestAccessToken.cs
```diff
@@ -6,6 +6,7 @@
 using System.Threading.Tasks;
 using D2L.Security.OAuth2.Provisioning;
 using D2L.Security.OAuth2.Scopes;
+using D2L.Security.OAuth2.Utilities;
```
omsmith, line 9: nit: some more no-longer-used usings left over

src/D2L.Security.OAuth2/Keys/Default/Data/IJwksProvider.cs
```diff
@@ -1,8 +1,11 @@
-﻿using System.Threading.Tasks;
+﻿using D2L.CodeStyle.Annotations;
+using System.Threading.Tasks;
```
omsmith, line 2: nit: convention is to place System usings first, and then the rest. It should be the Visual Studio default. There's a handful of occurrences.

image

src/D2L.Security.OAuth2/Keys/Default/Data/JwksProvider.cs
```diff
@@ -1,14 +1,17 @@
 ﻿using System;
+using System.IO;
```
omsmith, line 2: nit: unused

test/D2L.Security.OAuth2.Benchmarks/FullStackValidation/FullStackValidationBenchmark.cs
```diff
@@ -12,6 +13,7 @@ namespace D2L.Security.OAuth2.Benchmarks.FullStackValidation {
 	internal abstract class FullStackValidationBenchmark : IBenchmark {
 
 		private HttpClient m_httpClient;
+		private D2LHttpClient m_D2LHttpClient;
```
omsmith, line 16: Appears unused, given it's the only addition left in the file

j3parker (approved):
I think we should ultimately wrap HttpClient in something that implements the sync workflows with HttpWebRequest. The interface for that wrapper would be pretty "simple", e.g. it wouldn't return `HttpResponseMessage` (that object is too complex) and tailored to our (limited needs).

If we do that we maybe don't need the `Task.Run` wrapping support in the generator, which would be ideal.

I don't think we need to hold up this PR on any of that though, and I'm keen to see some e2e stuff in the LMS.

PR #314, opened 2023-07-10
Refactor of key expiry async stuff

j3parker:
This is an alternative implementation that is maybe simpler to transform into sync code. Not sure yet though.

### DIFF
--- a/src/D2L.Security.OAuth2/Keys/Default/ExpiringPublicKeyDataProvider.cs
+++ b/src/D2L.Security.OAuth2/Keys/Default/ExpiringPublicKeyDataProvider.cs
@@ -49,17 +49,21 @@ namespace D2L.Security.OAuth2.Keys.Default {
 				.GetAllAsync()
 				.ConfigureAwait( false );
 
-			keys = await Task
-				.WhenAll(
-					keys.Select( key => KeyExpirationHelper( key ) ).ToArray()
-				)
-				.ConfigureAwait( false );
+			var deletes = new List<Task>();
 
-			keys = keys
-				.Where( key => key != null )
-				.ToArray();
+			var result = new List<JsonWebKey );
 
-			return keys;
+			foreach( var key in keys ) {
+				var keyIsExpired = await KeyIsExpiredAsync( key ).ConfigureAwait( false );
+				
+				if( !keyIsExpired ) {
+					result.Add( key );
+				};
+			}
+
+			await Task.WhenAll( deletes ).ConfigureAwait( false );
+
+			return result;
 		}
 
 		Task IPublicKeyDataProvider.SaveAsync( Guid id, JsonWebKey key ) {
@@ -70,9 +74,9 @@ namespace D2L.Security.OAuth2.Keys.Default {
 			return m_inner.DeleteAsync( id );
 		}
 
-		private async Task<JsonWebKey> KeyExpirationHelper( JsonWebKey key ) {
+		private async Task<bool> KeyIsExpiredAsync( JsonWebKey key ) {
 			if( key == null ) {
-				return null;
+				return true;
 			}
 
 			if( key.ExpiresAt == null ) {
@@ -85,10 +89,10 @@ namespace D2L.Security.OAuth2.Keys.Default {
 				await ( this as IPublicKeyDataProvider )
 					.DeleteAsync( new Guid( key.Id ) )
 					.ConfigureAwait( false );
-				return null;
+				return true;
 			}
 
-			return key;
+			return false;
 		}
 	}
 }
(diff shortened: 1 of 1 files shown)


src/D2L.Security.OAuth2/Keys/Default/ExpiringPublicKeyDataProvider.cs
```diff
@@ -70,9 +74,9 @@ IDateTimeProvider dateTimeProvider
 			return m_inner.DeleteAsync( id );
 		}
 
-		private async Task<JsonWebKey> KeyExpirationHelper( JsonWebKey key ) {
+		private async Task<bool> KeyIsExpiredAsync( JsonWebKey key ) {
 			if( key == null ) {
-				return null;
+				return true;
```
j3parker, line 79: I'm not really sure what this is about tbh

src/D2L.Security.OAuth2/Keys/Default/ExpiringPublicKeyDataProvider.cs
```diff
@@ -49,17 +49,21 @@ IDateTimeProvider dateTimeProvider
...
-				.ConfigureAwait( false );
+			var deletes = new List<Task>();
+
+			var result = new List<JsonWebKey>();
 
-			keys = keys
-				.Where( key => key != null )
-				.ToArray();
+			foreach( var key in keys ) {
+				var keyIsExpired = await KeyIsExpiredAsync( key ).ConfigureAwait( false );
```
j3parker, line 57: 1. We need this to be named -Async for the sync-to-async generator
2. We want to not call blocking methods inside `.Select`/lambdas, because its hard to keep track of, so this PR eliminates the LINQ usage.

src/D2L.Security.OAuth2/Keys/Default/ExpiringPublicKeyDataProvider.cs
```diff
@@ -49,17 +49,17 @@ IDateTimeProvider dateTimeProvider
 				.GetAllAsync()
 				.ConfigureAwait( false );
 
-			keys = await Task
-				.WhenAll(
-					keys.Select( key => KeyExpirationHelper( key ) ).ToArray()
-				)
```
j3parker, line 55: We can transform WhenAll but there is some trickiness.
The `[Blocking]` analyzer doesn't allow calls to blocking methods (like a syncified `KeyExpirationHelper`) inside Lambdas because it would be too much to keep track of.

This PR removes the WhenAll (in practice we should typically have 0 and sometimes 1 task to await, so concurrency doesn't matter) and the lambda.

AnaCoda (approved):

src/D2L.Security.OAuth2/Keys/Default/ExpiringPublicKeyDataProvider.cs
omsmith: For a particular client, here's the current relative expiry in seconds of each of their keys from the earliest expiring key.

```
[
     0,    0,    1,    8,   12,   62,
    63,   63,   66, 1572, 2312, 2406,
  2604, 2704, 2826, 2943, 2956, 2969,
  2970, 3099, 3601, 3603, 3603, 3610,
  3613, 3663, 3665, 3665, 3667
]
```

I think I given call to them might easily capture the first 9 of them to expire?

omsmith: Or represented as distance from the next oldest key:

```
> expiries.sort().map((x, i) => x-(i==0?x:expiries[i-1]))
[
    0,   0,    1,   7,  4,  50,   1,
    0,   3, 1506, 740, 94, 198, 100,
  122, 117,   13,  13,  1, 129, 502,
    2,   0,    7,   3, 50,   2,   0,
    2
]
```

and another client:

```
[
    0, 238, 40, 48, 379, 112,  7, 311, 404,  29,
  152,   0,  0, 49,   3,   8, 19,  34,   1,   3,
    8,  55,  3, 19,  41,   6,  1,   5,  27,   2,
   17,   1, 16, 38, 287,  30, 31,  44, 258, 468,
  411, 267
]
```

Probably still fine, just want to be aware I think this will be 2+ in many cases as well.

omsmith (approved):

src/D2L.Security.OAuth2/Keys/Default/ExpiringPublicKeyDataProvider.cs
j3parker: Oh ya I forgot about the LMS and its key craziness.

PR #324, opened 2023-08-18
Generate sync AccessTokenProvider

AnaCoda:


src/D2L.Security.OAuth2/Caching/NullCache.cs
```diff
@@ -1,15 +1,25 @@
...
+		[GenerateSync]
+		Task ICache.SetAsync( string key, string value, TimeSpan expiry ) {
+			return Task.CompletedTask;
+		}
 
-		Task ICache.RemoveAsync( string key ) => Task.CompletedTask;
+		[GenerateSync]
+		Task ICache.RemoveAsync( string key ) {
+			return Task.CompletedTask;
+		}
```
j3parker, line 23: For my own amusement can you post the generated code for this file?

j3parker (approved):

omsmith (changes requested):
Just want to mark that I'd like to review the generated code locally before this merges.

src/D2L.Security.OAuth2/Caching/NullCache.cs
omsmith:
```csharp
#pragma warning disable CS1572
using System;
using System.Threading.Tasks;
using D2L.CodeStyle.Annotations;

namespace D2L.Security.OAuth2.Caching {
	partial class NullCache {

		[Blocking]
		CacheResponse ICache.Get( string key ) {
			return ( NULL_RESPONSE );
		}

		[Blocking]
		void ICache.Set( string key, string value, TimeSpan expiry ) {
			return;
		}

		[Blocking]
		void ICache.Remove( string key ) {
			return;
		}
	}
}
```

AnaCoda:
```
namespace D2L.Security.OAuth2.Caching {
	partial class NullCache {

		[Blocking]
		CacheResponse ICache.Get( string key ) {
			return ( NULL_RESPONSE );
		}

		[Blocking]
		void ICache.Set( string key, string value, TimeSpan expiry ) {
			return;
		}

		[Blocking]
		void ICache.Remove( string key ) {
			return;
		}
	}
}
```

src/D2L.Security.OAuth2/Keys/Default/RotatingPrivateKeyProvider.cs
```diff
@@ -25,6 +26,7 @@ TimeSpan keyRotationPeriod
 			m_keyRotationPeriod = keyRotationPeriod;
 		}
 
+		[GenerateSync]
 		async Task<D2LSecurityToken> IPrivateKeyProvider.GetSigningCredentialsAsync() {
```
omsmith, line 30: This function loses a critical `try` because it only has a `finally` block

```csharp
		[Blocking]
D2LSecurityToken IPrivateKeyProvider.GetSigningCredentials() {

			// Hold a local reference so that we know we are talking about the same key
			// after even if another thread changed m_privateKey (race condition when we
			// are using a key very close to the rotation time.)
			D2LSecurityToken privateKey = m_privateKey;

			if( NeedFreshPrivateKey( privateKey ) ) {
				// This Semaphore is used instead of lock(foo){}
				// because await cannot be used within a lock
				m_privateKeyLock.Wait();
{
					privateKey = m_privateKey;

					if( NeedFreshPrivateKey( privateKey ) ) {
						m_privateKey = ( m_inner.GetSigningCredentials()).Ref();

						if( privateKey != null ) {
							privateKey.Dispose();
						}

						privateKey = m_privateKey;
					}

				} 			}

			return privateKey.Ref();
		}
```

AnaCoda: Oh, very good catch! I will fix that in the generator

src/D2L.Security.OAuth2/Keys/Default/TokenSigner.cs
```diff
@@ -19,6 +20,7 @@ IPrivateKeyProvider privateKeyProvider
 			m_privateKeyProvider = privateKeyProvider;
 		}
 
+		[GenerateSync]
 		async Task<string> ITokenSigner.SignAsync( UnsignedToken token ) {
 			JwtSecurityToken jwt;
 			using( D2LSecurityToken securityToken = await m_privateKeyProvider
```
omsmith, line 26: Doesn't matter, but the `JwtSecurityToken` `ObjectCreationExpression` has a curious loss of newlines:

```csharp
				jwt = new JwtSecurityToken(
					issuer: token.Issuer,					audience: token.Audience,					claims: Enumerable.Empty<Claim>(),					notBefore: token.NotBefore,					expires: token.ExpiresAt,					signingCredentials: securityToken.GetSigningCredentials()
				);
```

omsmith: Probably has to do with commas in invocations losing their trivia in general

`m_cache.Get( m_jwksProvider.Namespace, id ); -> m_cache.Get( m_jwksProvider.Namespace,id );`

src/D2L.Security.OAuth2/Keys/Default/RotatingPrivateKeyProvider.cs
j3parker: Spicy! Nice catch.

src/D2L.Security.OAuth2/Caching/NullCache.cs
j3parker: awesome

src/D2L.Security.OAuth2/Keys/Default/RsaPrivateKeyProvider.cs
```diff
@@ -14,6 +15,7 @@ ID2LSecurityTokenFactory d2lSecurityTokenFactory
 			m_d2lSecurityTokenFactory = d2lSecurityTokenFactory;
 		}
 
+		[GenerateSync]
 		Task<D2LSecurityToken> IPrivateKeyProvider.GetSigningCredentialsAsync() {
```
omsmith, line 19: For not-actually-async functions like this; I wonder if we'd ultimately want to do something different?

omsmith (approved):
Reading over the generated code, everything is looking okay. Some of it is a little trickier to read due to whitespace issues. Annnd yeah, lets just be real careful, that missing `try` was scary 😄

src/D2L.Security.OAuth2/Keys/Default/RsaPrivateKeyProvider.cs
AnaCoda: @j3parker

omsmith: I understand that, yes.

PR #342, opened 2023-11-21
Adding 'iat' (issued at) claim to tokens

joshhoey:
Adding `iat` (Issued At) as a claim (same as `nbf` for now, but it means something different)

### DIFF
--- a/src/D2L.Security.OAuth2/Keys/Default/TokenSigner.cs
+++ b/src/D2L.Security.OAuth2/Keys/Default/TokenSigner.cs
@@ -1,4 +1,5 @@
-﻿using System.IdentityModel.Tokens.Jwt;
+﻿using System;
+using System.IdentityModel.Tokens.Jwt;
 using System.Linq;
 using System.Security.Claims;
 using System.Threading.Tasks;
@@ -35,6 +36,8 @@ namespace D2L.Security.OAuth2.Keys.Default {
 					expires: token.ExpiresAt,
 					signingCredentials: securityToken.GetSigningCredentials()
 				);
+
+				jwt.Payload.Add(OAuth2.Constants.Claims.ISSUED_AT, ((DateTimeOffset)token.IssuedAt).ToUnixTimeSeconds());
 
 				var claims = token.Claims;
 				foreach( var claim in claims ) {
--- a/src/D2L.Security.OAuth2/Keys/Default/TokenSigner.cs
+++ b/src/D2L.Security.OAuth2/Keys/Default/TokenSigner.cs
@@ -37,7 +37,8 @@ namespace D2L.Security.OAuth2.Keys.Default {
 					signingCredentials: securityToken.GetSigningCredentials()
 				);
 
-				jwt.Payload.Add(OAuth2.Constants.Claims.ISSUED_AT, ((DateTimeOffset)token.IssuedAt).ToUnixTimeSeconds());
+				DateTimeOffset issuedAtOffset = token.IssuedAt;
+				jwt.Payload.Add( OAuth2.Constants.Claims.ISSUED_AT, issuedAtOffset.ToUnixTimeSeconds() );
 
 				var claims = token.Claims;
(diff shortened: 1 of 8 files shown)


src/D2L.Security.OAuth2/Keys/UnsignedToken.cs
```diff
@@ -69,5 +72,14 @@ DateTime expiresAt
 		public DateTime ExpiresAt {
 			get { return m_expiresAt; }
 		}
+
+		/// <summary>
+		/// When the token was issued; this is the 'iat' claim
+		/// </summary>
+		public DateTime IssuedAt
+		{
```
omsmith, line 80: nit:

```suggestion
		public DateTime IssuedAt {
```

src/D2L.Security.OAuth2/Keys/Default/TokenSigner.cs
```diff
@@ -36,6 +37,8 @@ IPrivateKeyProvider privateKeyProvider
 					signingCredentials: securityToken.GetSigningCredentials()
 				);
 
+				jwt.Payload.Add(OAuth2.Constants.Claims.ISSUED_AT, ((DateTimeOffset)token.IssuedAt).ToUnixTimeSeconds());
```
omsmith, line 40: Would prefer to use the implicit constructor vs an explicit cast. Explicit casts can hide conversions that are never valid.

Also nit on whitespace
```suggestion
				DateTimeOffset issuedAtOffset = token.IssuedAt;
				jwt.Payload.Add( OAuth2.Constants.Claims.ISSUED_AT, issuedAtOffset.ToUnixTimeSeconds() );
```

omsmith (changes requested):

omsmith (approved):

j3parker:
Hm, I thought we had iat but we have nbf and exp. I wonder if we could just get by with those? (probably just nbf?)

joshhoey:
> Hm, I thought we had iat but we have nbf and exp. I wonder if we could just get by with those? (probably just nbf?)

Yeah, we could get by with nbf (it would have made my life a bit easier) but nbf doesn't strictly need to match iat, we're just making it "now" by convention, so at this point we figured we'd keep going and add it. 🤷‍♀️

omsmith (approved):

j3parker:
> Given the semantic difference, I preferred adding the proper claim.

We can also just use one if it exists and if not the other. iat isn't a required claim.

j3parker (changes requested):
We don't need more claims

omsmith:
I don't think it being required or not is a particularly compelling reason not to use it. _nbf_ and _exp_ are also optional claims, but we of course use them due to the semantics of what they imply (and require of client libraries).

We do not currently have any usage of _nbf_ being set to a future date, and I don't think it's _terribly_ likely that we would end up wanting to either. However I'm not sure how to assert that we never do so such that the general-we would catch the change down the line.

I was definitely iffy about adding extra bytes to the tokens, but the semantic of _iat_ is exactly what we're looking for, so suggested it as being more appropriate despite that.

Can you think of a way to future-proof ourselves against future mistakes re _nbf_ being adjusted away from time-of-issue, and thus failing to revoke those tokens?

j3parker:
> Can you think of a way to future-proof ourselves against future mistakes re nbf being adjusted away from time-of-issue, and thus failing to revoke those tokens?

```cs
var threshold = token.IssuedAt ?? token.NotBefore;
```

is what I meant by "we can also just use one if it exists and if not the other".
Like we don't have iat right now because it would mostly be redundant, so I think it's fine to just do that.

j3parker:
I would also not be opposed to `min(nbf, iat)` to be conservative. I don't think there are any rules about which of those come earlier, and using the earlier of the two would always revoke a greater number of tokens which I think is what we'd want (assuming we're putting reasonable values into iat/nbf and not like "0" -- JWT claims like this have barely any semantics/its left to implementations to decide this stuff.)

PR #359, opened 2024-02-07
Have GetSigningCredentialsAsync refresh its key earlier

j3parker:
Only refreshing after the key is expired will cause us to sign tokens that might not validate for their entire lifetime.

The spirit was to have GetSigningCredentialsAsync call RefreshAsync even if there was no background job calling it -- so this change makes it call it at the same cadence/time we expect a background service to.

This hasn't been a problem in production for various reasons but it was for a specific use-case in the dev environment, but regardless its a bug.

### DIFF
--- a/src/D2L.Security.OAuth2/Keys/KeyManagementService.cs
+++ b/src/D2L.Security.OAuth2/Keys/KeyManagementService.cs
@@ -51,7 +51,7 @@ namespace D2L.Security.OAuth2.Keys {
 
 			var now = m_clock.UtcNow;
 
-			if ( current == null || current.ValidTo <= now ) {
+			if ( current == null || ExpectedTimeOfNewUsableKey( current ) < now ) {
 				// Slow path: RefreshKeyAsync() wasn't called on boot and/or it
 				// isn't being called in a background job.
 				await RefreshKeyAsync( now )
@@ -67,6 +67,13 @@ namespace D2L.Security.OAuth2.Keys {
 			return current.Ref();
 		}
 
+		private DateTimeOffset ExpectedTimeOfNewUsableKey( D2LSecurityToken current )
+			// A new key will get generated some time before the current key
+			// expires, but will only become usable some time after that.
+			=> current.ValidTo
+				- m_config.KeyRotationBuffer
+				+ m_config.KeyTimeUntilUse;
+
 		[GenerateSync]
 		async Task<TimeSpan> IKeyManagementService.RefreshKeyAsync() {
 			var now = m_clock.UtcNow;
@@ -81,11 +88,7 @@ namespace D2L.Security.OAuth2.Keys {
 				return TimeSpan.FromSeconds( 10 );
 			}
 
-			// A new key will get generated some time before the current key
-			// expires, but will only become usable some time after that.
-			var expectedTimeOfNewUsableKey = current.ValidTo
-				- m_config.KeyRotationBuffer
-				+ m_config.KeyTimeUntilUse;
+			var expectedTimeOfNewUsableKey = ExpectedTimeOfNewUsableKey( current );
 
 			if( now > expectedTimeOfNewUsableKey ) {
 				// If we would have expected a new key by now, retry again in a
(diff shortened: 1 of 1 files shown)


src/D2L.Security.OAuth2/Keys/KeyManagementService.cs
```diff
@@ -51,7 +51,7 @@ OAuth2Configuration config
 
 			var now = m_clock.UtcNow;
 
-			if ( current == null || current.ValidTo <= now ) {
+			if ( current == null || ExpectedTimeOfNewUsableKey( current ) < now ) {
```
j3parker, line 54: Oh I guess we talked about adding a buffer here, one sec

src/D2L.Security.OAuth2/Keys/KeyManagementService.cs
```diff
@@ -17,6 +17,20 @@ public sealed partial class KeyManagementService : IKeyManagementService, IPriva
...
+
+		// This controls how frequently background refresh will retry if it
+		// doesn't find a key
+		private static readonly TimeSpan BackgroundRefreshRetryDelay = TimeSpan.FromMinutes( 1 );
+
+		// Wait for the delay and some number of retries before making GetSigningCredentials
+		// do a foreground Refresh
+		private static readonly TimeSpan GetSigningCredentialsRefreshGracePeriod
+			= BackgroundRefreshDelay
+			+ BackgroundRefreshRetryDelay + BackgroundRefreshRetryDelay;
```
j3parker, line 32: can't do `2*` 😒

src/D2L.Security.OAuth2/Keys/KeyManagementService.cs
```diff
@@ -51,7 +65,9 @@ OAuth2Configuration config
 
 			var now = m_clock.UtcNow;
 
-			if ( current == null || current.ValidTo <= now ) {
+			if ( current == null
+				|| ExpectedTimeOfNewUsableKey( current ) + GetSigningCredentialsRefreshGracePeriod < now
```
j3parker, line 69: The implication here is that once this period is passed every call to `GetSigningCredentials` is going to fight to update that key.

That basically matches how it worked before the background refresh anyway. The only things without background refresh for us are Lambda functions though, which are single threaded anyway.

src/D2L.Security.OAuth2/Keys/KeyManagementService.cs
j3parker: Done

daryl-mcmillan: If we were comfortable storing the refresh time somewhere the TimeSpan value returned by RefreshKeyAsync has a bunch of stuff built into it:
- if we have no valid key retry in 10 seconds
- retry 1 minute after a new key should be available
- retry every minute if there's supposed to be a newer key available

omsmith (approved):

src/D2L.Security.OAuth2/Keys/KeyManagementService.cs
```diff
@@ -51,7 +65,9 @@ OAuth2Configuration config
 
 			var now = m_clock.UtcNow;
 
-			if ( current == null || current.ValidTo <= now ) {
+			if ( current == null
+				|| ExpectedTimeOfNewUsableKey( current ) + GetSigningCredentialsRefreshGracePeriod < now
+			) {
 				// Slow path: RefreshKeyAsync() wasn't called on boot and/or it
 				// isn't being called in a background job.
 				await RefreshKeyAsync( now )
```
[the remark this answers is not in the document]
daryl-mcmillan: If key rotation fails for 3 minutes I think the results are pretty bad here?  Taking a look now at whether dev key rotation ever gets delayed...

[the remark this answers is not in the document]
daryl-mcmillan: dev dynamo metrics show writes every 3 hours in the first 5 minutes of the hour steady for the past two months.  I think we're still in pretty rough shape if cert generation gets delayed, but hopefully that won't happen too often.

daryl-mcmillan (approved):

PR #360, opened 2024-02-08
the key caching has some race conditions

daryl-mcmillan:
This is assuming the reference counting via `Dispose()` is reasonable - really multiple Dispose calls are supposed to be ok so `Ref()` should probably be returning a wrapper.
https://stackoverflow.com/a/5306896/24158

### DIFF
--- a/src/D2L.Security.OAuth2/Keys/KeyManagementService.cs
+++ b/src/D2L.Security.OAuth2/Keys/KeyManagementService.cs
@@ -5,7 +5,6 @@ using Microsoft.IdentityModel.Tokens;
 using System;
 using System.Collections.Generic;
 using System.Security.Cryptography;
-using System.Threading;
 using System.Threading.Tasks;
 
 namespace D2L.Security.OAuth2.Keys {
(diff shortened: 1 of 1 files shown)


src/D2L.Security.OAuth2/Keys/KeyManagementService.cs
```diff
@@ -57,14 +70,11 @@ OAuth2Configuration config
...
-				current = Volatile.Read( ref m_current );
+				using var refreshed = GetCurrentToken();
+				return refreshed?.Ref();
 			}
 
-			if( current == null ) {
-				return null;
-			}
-
-			return current.Ref();
```
daryl-mcmillan, line 67: this is the main race condition - if the key is replaced between `Volatile.Read` and this call to `.Ref()` is will be disposed.

src/D2L.Security.OAuth2/Keys/KeyManagementService.cs
```diff
@@ -15,8 +14,22 @@ public sealed partial class KeyManagementService : IKeyManagementService, IPriva
 		private readonly IDateTimeProvider m_clock;
 		private readonly OAuth2Configuration m_config;
 
+		private readonly object m_keyRefreshLock = new object();
 		private D2LSecurityToken m_current = null;
 
+		private D2LSecurityToken GetCurrentToken() {
+			lock( m_keyRefreshLock ) {
+				return m_current?.Ref();
```
daryl-mcmillan, line 22: don't let the key be disposed while we're thinking about returning it.

PR #364, opened 2024-07-31
LFT-1285 - Ignore use=enc JWKs in JWKS parsing

j3parker:
We never have use=enc keys and so our parsing throws when it encounters them.

We don't distinguish between unparsable keys vs. keys that we don't want to use for signing (e.g. use=enc, but also other constraints we impose like having a suitable kty, kid etc.)

In an ideal world I think we would push more of that logic out a bit and parse more keys, ignoring ones that aren't relevant to us anyway and emitting errors in a more sensible location.

This PR makes the minimal change possible to allow us to parse a JWKS response that includes a use=enc key that we will otherwise not need. This is needed for interoperability with LTI vendors that have a JWKS URL serving multiple purposes.


src/D2L.Security.OAuth2/Keys/Default/JsonWebKeySet.cs
```diff
@@ -20,10 +20,22 @@ internal sealed class JsonWebKeySet {
...
+				var builder = ImmutableArray.CreateBuilder<JsonWebKey>(
+					initialCapacity: keyObjects.Count
+				);
+
 				foreach( object keyObject in keyObjects ) {
 					string keyJson = JsonConvert.SerializeObject( keyObject );
-					JsonWebKey key = JsonWebKey.FromJson( keyJson );
+
+					if( !JsonWebKey.TryParseJsonWebKey( keyJson, out var key, out var error, out var exception, out var useEncKey ) ) {
+						if( useEncKey ) {
```
j3parker, line 31: This filtering could apply more generally but for now I just want to do the minimal thing to get the integrations working. I think the API needs a redesign overall.

src/D2L.Security.OAuth2/Keys/JsonWebKey.cs
```diff
@@ -49,91 +51,169 @@ public abstract class JsonWebKey {
...
+		public static bool TryParseJsonSigningKey(
+			string json,
+			[NotNullWhen( true )]
+			out JsonWebKey? key,
+			[NotNullWhen( false )]
+			out string? error,
+			[NullWhen( true )]
+			out Exception? exception,
+			out bool useEncKey
+		) {
```
j3parker, line 77: Mostly changed to TryX from throwing, but also added this separate `useEncKey` signal.

src/D2L.Security.OAuth2/NotNullWhenAttribute.net4x.cs
```diff
@@ -0,0 +1,10 @@
+namespace System.Diagnostics.CodeAnalysis;
```
j3parker, line 1: (Note filepath: .net4x.cs is treated specially via our build.)

src/D2L.Security.OAuth2/Keys/JsonWebKey.cs
```diff
@@ -60,81 +63,153 @@ public abstract class JsonWebKey {
...
 
 			if( data.ContainsKey( "use" ) && data[ "use" ] != null && data[ "use" ].ToString() != "sig" ) {
-				string msg = String.Format( "invalid 'use' value in JSON web key: {0}", data[ "use" ] );
-				throw new JsonWebKeyParseException( msg );
+				result = null;
+				error = "invalid 'use' value in JSON web key: " + data[ "use" ];
+				exception = null;
+				useEncKey = data[ "use" ].ToString() == "enc";
+				return false;
 			}
```
omsmith, line 99: I would maybe make this ignored in general (and not just "enc")? Perhaps a `ParseResult { Success, Error, Ignore }` vs `bool`?

It would ultimately be nice to be able to say "the key {kid} is not intended for singnatures" rather than "the key {kid} was not found" though.

omsmith (approved):
Seems fine enough

src/D2L.Security.OAuth2/Keys/JsonWebKey.cs
j3parker: In other words rather than returning ignore from here I'd prefer if we just successfully parsed all valid JWKs and then moved the policy decisions elsewhere

PR #366, opened 2024-08-01
LFT-1285 - Suppress sync generator in .NET Core build

j3parker:
We don't need the sync implementations but we especially don't want the obligations to implement sync APIs in .NET Core. The spirit of the generator is to make migrating to sync in .NET Framework sustainable, the idea generally is that in .NET Core we won't rely on sync APIs.

### DIFF
--- a/src/D2L.Security.OAuth2/D2L.Security.OAuth2.csproj
+++ b/src/D2L.Security.OAuth2/D2L.Security.OAuth2.csproj
@@ -10,6 +10,10 @@
 		<IncludeSymbols>true</IncludeSymbols>
 	</PropertyGroup>
 
+  <ItemGroup>
+    <CompilerVisibleProperty Include="SuppressSyncGenerator" />
+  </ItemGroup>
+
 	<ItemGroup>
 		<PackageReference Include="D2L.CodeStyle.Analyzers">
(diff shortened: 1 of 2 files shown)


j3parker:
Built locally and compared the DLLs in ILSpy; `IPublicKeyDataProvider` does not have sync methods in the net60 build: 
image

j3parker:
Here's ILSpy showing the net60 build (top) without sync overloads in IPublicKeyDataProvider, and the net472 build (below) with them.

image

src/D2L.Security.OAuth2/D2L.Security.OAuth2.csproj
```diff
@@ -10,6 +10,10 @@
 		<IncludeSymbols>true</IncludeSymbols>
 	</PropertyGroup>
 
+  <ItemGroup>
+    <CompilerVisibleProperty Include="SuppressSyncGenerator" />
+  </ItemGroup>
```
j3parker, line 15: Hm, I'll read up on the docs, I was trying to make it so this was not necessary here. I'll probably need another bump of analyzers to get rid of this now.

omsmith (approved):

PR #373, opened 2024-11-29
Move caching stuff off the IAccessTokenProvider API into the factory

j3parker:


src/D2L.Security.OAuth2.TestFramework/TestAccessTokenProviderFactory.cs
```diff
@@ -38,9 +39,9 @@ public static IAccessTokenProvider Create( HttpClient httpClient, String tokenPr
 			Uri authEndpoint = new Uri( tokenProvisioningEndpoint );
 			ITokenSigner tokenSigner = new TokenSigner( privateKeyProvider );
 			IAuthServiceClient authServiceClient = new AuthServiceClient( httpClient, authEndpoint );
-			INonCachingAccessTokenProvider noCacheTokenProvider = new AccessTokenProvider( tokenSigner, authServiceClient );
+			IAccessTokenProvider noCacheTokenProvider = new AccessTokenProvider( tokenSigner, authServiceClient );
 
-			return new CachedAccessTokenProvider( noCacheTokenProvider, authEndpoint, Timeout.InfiniteTimeSpan );
+			return new CachedAccessTokenProvider( new NullCache(), noCacheTokenProvider, authEndpoint, Timeout.InfiniteTimeSpan );
```
j3parker, line 44: tests

src/D2L.Security.OAuth2/Provisioning/AccessTokenProviderFactory.cs
```diff
@@ -18,18 +19,34 @@ public static IAccessTokenProvider Create(
 			ITokenSigner tokenSigner,
 			HttpClient httpClient,
 			Uri authEndpoint,
-			TimeSpan tokenRefreshGracePeriod
+			TimeSpan tokenRefreshGracePeriod,
+			IAccessTokenProvider inner = null,
+			ICache cache = null
 		) {
+			if( inner != null && cache == null ) {
+				throw new InvalidOperationException( "If you provide an inner you need to also provide its cache" );
```
j3parker, line 27: The idea here is that this factory is either for wrapping your IAccessTokenProvider in our caching layer (for which you provide an ICache for us to use) or for using our jwt-bearer-grant based impl with or without caching.

The reason I'm doing this is I want to implement an IamGrant IAccessTokenProvider impl. I want to do that inside the LMS because I'd really like to use the LMS Lambda framework. It would make sense to implement it in this library, and I mean you could at least define the input/output here, but meh.

src/D2L.Security.OAuth2/Provisioning/IAccessTokenProvider.cs
```diff
@@ -16,15 +15,13 @@ public partial interface IAccessTokenProvider {
...
-		/// <param name="cache">The provided <see cref="ICache"/> does not need to 
 		/// check for token expiration or grace period because the 
 		/// <see cref="IAccessTokenProvider"/> will handle it internally.</param>
 		/// <returns>An access token containing an expiry and the provided claims and scopes.</returns>
 		[GenerateSync]		
 		Task<IAccessToken> ProvisionAccessTokenAsync(
 			IEnumerable<Claim> claims,
-			IEnumerable<Scope> scopes,
-			ICache cache = null
+			IEnumerable<Scope> scopes
```
j3parker, line 24: Consolidating on one interface, equivalent to the old non-caching one, and taking the cache inside the factory+constructor.

The LMS cache is instantiated each call but it can be safely cached.

omsmith (approved):

PR #374, opened 2025-05-08
LFT-1689 - Remove Newtonsoft from the .NET 4 build of D2L.Security.OAuth2

j3parker:
It remains in place in some of the test DLLs for now.
I think we can remove it there but it is less important, I'd like to see this work first.


src/D2L.Security.OAuth2.TestFramework/AuthServiceMock.cs
```diff
@@ -2,6 +2,7 @@
 using System.Collections.Generic;
 using System.Net.Http;
 using System.Security.Cryptography;
+using System.Text.Json;
```
j3parker, line 5: The changes in tests here to add the using are because the JsonSerializer helper that we defined in the main package is now gone.
That means these tests are now using both System.Text.Json and Newtonsoft which is a bit ugly but I'm not worried about it.

If everything goes well I'll delete the references to Newtonsoft in the tests too.

.github/workflows/ci.yml
```diff
@@ -19,6 +19,10 @@ jobs:
     steps:
       - uses: Brightspace/third-party-actions@actions/checkout
       - uses: Brightspace/third-party-actions@actions/setup-dotnet
+        with:
+          dotnet-version: |
+            6.0.x
+            7.0.x
```
j3parker, line 25: This codifies the previous expectations of the repo explicitly.
The runner OS images have updated and what versions of .NET are installed has changed, so the implicit assumptions became broken.

We could probably upgrade this package to update something more recent than .NET 6 but realistically there's not much point so for now I'm just maintaining the status quo.

daryl-mcmillan (approved):

PR #380, opened 2026-03-31
NEX-5068 - Ensure that .NET 6 is installed in Windows build too

j3parker:


.github/workflows/ci.yml
```diff
@@ -37,6 +37,10 @@ jobs:
     steps:
       - uses: Brightspace/third-party-actions@actions/checkout
       - uses: Brightspace/third-party-actions@actions/setup-dotnet
+       with:
+          dotnet-version: |
+            6.0.x
+            7.0.x
```
j3parker, line 43: This matches the Linux build. We should probably update our .NET version but it's not really important.
This is required to get a package version published.

mahmoudialireza (approved):

PR #413, opened 2026-05-26
LFT-2890: Upgrade to .NET 10

j3parker:
Motivated by wanting to add nullable annotations to this library. The stdlib has gotten slightly better annotations since .NET 6 (I noticed this while adding my own)

.NET 8 and .NET 10 are both LTS versions, and .NET 9 is currently supported, but I couldn't think of a compelling motivation to not require .NET 10 going forward.

Aside: I will not ship nullable annotations with .NET 4.7.2 because it isn't well supported there.


src/D2L.Security.OAuth2.TestFramework/D2L.Security.OAuth2.TestFramework.csproj
```diff
@@ -14,7 +14,7 @@
 		<PackageReference Include="RichardSzalay.MockHttp" />
 	</ItemGroup>
 
-	<ItemGroup Condition="'$(TargetFramework)' != 'net60'">
+	<ItemGroup Condition="'$(TargetFramework)' == 'net472'">
```
j3parker, line 17: I flipped all the conditions to look at net472 because that one ought to be more stable

Directory.Build.targets
```diff
@@ -1,11 +1,11 @@
...
 		<Compile Remove="$(MSBuildProjectDirectory)/**/*.net4x.cs" />
 		<None Include="$(MSBuildProjectDirectory)/**/*.net4x.cs" />
 	</ItemGroup>
 
-	<ItemGroup Condition="'$(TargetFramework)' != 'net60'">
-		<Compile Remove="$(MSBuildProjectDirectory)/**/*.net60.cs" />
-		<None Include="$(MSBuildProjectDirectory)/**/*.net60.cs" />
+	<ItemGroup Condition="'$(TargetFramework)' == 'net472'">
+		<Compile Remove="$(MSBuildProjectDirectory)/**/*.net5plus.cs" />
+		<None Include="$(MSBuildProjectDirectory)/**/*.net5plus.cs" />
```
j3parker, line 9: Changed filename suffix to be more stable. There is currently only one of these.

omsmith (approved):

PR #414, opened 2026-05-26
LFT-2890: Use TryGetValue where appropriate

j3parker:
It's faster to not look stuff up twice but it also plays better with nullable analysis

### DIFF
--- a/src/D2L.Security.OAuth2/Keys/JsonWebKey.cs
+++ b/src/D2L.Security.OAuth2/Keys/JsonWebKey.cs
@@ -3,6 +3,7 @@
 using System;
 using System.Collections.Generic;
 using System.Diagnostics.CodeAnalysis;
+using System.Security.Cryptography;
 using System.Text.Json;
 using D2L.Security.OAuth2.Keys.Default;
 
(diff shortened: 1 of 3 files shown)


src/D2L.Security.OAuth2/Keys/JsonWebKey.cs
j3parker, line 82: the null diagnostic here is mostly legit but pre-existing; I will be addressing that in a future PR (my goal is to enable nullable everywhere + make them errors)

src/D2L.Security.OAuth2/Keys/JsonWebKey.cs
```diff
@@ -86,54 +87,54 @@ out bool useEncKey
 				return false;
 			}
 
-			if( data.ContainsKey( "use" ) && data[ "use" ] != null && data[ "use" ].ToString() != "sig" ) {
+			if( data.TryGetValue( "use", out var use ) && use != null && use.ToString() != "sig" ) {
 				result = null;
```
j3parker, line 91: Hm, the null warning here is on `.ToString()`, because the claim values are `object` and the return type of that is `string?`. These are pre-existing diagnostics.

That return type is sort of a historical quirk/design choice; it has always been part of the guidelines that you are not to return `null` from `ToString()` because that's dumb, but `object` is so fundamental that even within the standard library there was some non-compliance.

Narrower types can use `string` as their return which is why this often doesn't come up.

In a future PR I added `!` to these `.ToString()` along with a comment explaining why I think that's OK. If it's OK T

omsmith (approved):

PR #415, opened 2026-05-27
LFT-2890: Enable Nullable errors in D2L.Security.OAuth2

j3parker:


src/D2L.Security.OAuth2/Keys/Default/EcDsaPrivateKeyProvider.cs
```diff
@@ -19,17 +19,17 @@ ECCurve curve
...
+		Task<D2LSecurityToken?> IPrivateKeyProvider.GetSigningCredentialsAsync() {
 			var ecdsa = ECDsa.Create( m_curve );
 			var parameters = ecdsa.ExportParameters( includePrivateParameters: true );
 
-			D2LSecurityToken result = m_d2lSecurityTokenFactory.Create( () => {
+			var result = m_d2lSecurityTokenFactory.Create( () => {
 				var ecDsa = ECDsa.Create( parameters );
 				var key = new ECDsaSecurityKey( ecDsa );
 				return new Tuple<AsymmetricSecurityKey, IDisposable>( key, ecDsa );
 			} );
```
j3parker, line 30: note: when something could be solved with `var` that is what I did.

src/D2L.Security.OAuth2/Keys/Default/JsonWebKeySet.cs
```diff
@@ -15,6 +16,10 @@ public JsonWebKeySet( string json, Uri src ) {
 			try {
 				var data = JsonSerializer.Deserialize<Dictionary<string, object>>( json );
 
+				if( data == null ) {
+					throw new JsonWebKeyParseException( "invalid json web key set: could not deserialize json" );
+				}
```
j3parker, line 21: Behavious change: if `json` is for example `"null"` this would deserialize as `null` and then we would throw a `NullReferenceException` below. This seemed better.

src/D2L.Security.OAuth2/Keys/Default/RotatingPrivateKeyProvider.cs
```diff
@@ -42,7 +42,7 @@ async Task<D2LSecurityToken> IPrivateKeyProvider.GetSigningCredentialsAsync() {
 					privateKey = m_privateKey;
 
 					if( NeedFreshPrivateKey( privateKey ) ) {
-						m_privateKey = ( await m_inner.GetSigningCredentialsAsync().ConfigureAwait( false ) ).Ref();
+						m_privateKey = ( await m_inner.GetSigningCredentialsAsync().ConfigureAwait( false ) )?.Ref();
```
j3parker, line 45: Behaviour change, we used to throw when there was no private key, now we don't and instead return `null` but let the client decide what to do.

This is a rare edge case, like it has never actually happened and shouldn't, but I think it is reasonable to ask the caller to figure out what they want to do.

src/D2L.Security.OAuth2/Keys/Default/RotatingPrivateKeyProvider.cs
```diff
@@ -56,10 +56,10 @@ async Task<D2LSecurityToken> IPrivateKeyProvider.GetSigningCredentialsAsync() {
 				}
 			}
 
-			return privateKey.Ref();
+			return privateKey?.Ref();
 		}
 
-		private bool NeedFreshPrivateKey( D2LSecurityToken key ) {
+		private bool NeedFreshPrivateKey( [NotNullWhen( false )] D2LSecurityToken? key ) {
```
j3parker, line 62: Rare `[NotNullWhen( false )]` lol

src/D2L.Security.OAuth2/Keys/Default/SavingPrivateKeyProvider.cs
```diff
@@ -17,12 +17,14 @@ ISanePublicKeyDataProvider publicKeyDataProvider
...
+		async Task<D2LSecurityToken?> IPrivateKeyProvider.GetSigningCredentialsAsync() {
+			var result = await m_inner.GetSigningCredentialsAsync().ConfigureAwait( false );
 
-			JsonWebKey jwk = result.ToJsonWebKey();
 
-			await m_publicKeyDataProvider.SaveAsync( new Guid( jwk.Id ), jwk ).ConfigureAwait( false );
+			if( result != null) {
+				var jwk = result.ToJsonWebKey();
+				await m_publicKeyDataProvider.SaveAsync( new Guid( jwk.Id ), jwk ).ConfigureAwait( false );
+			}
```
j3parker, line 27: Behaviour change (returning `null` vs immediately throwing) but related to the previously discussed contract change; this API is up-front that it might return `null`

src/D2L.Security.OAuth2/Keys/Default/TokenSigner.cs
```diff
@@ -33,7 +33,7 @@ async Task<string> ITokenSigner.SignAsync( UnsignedToken token ) {
 					claims: Enumerable.Empty<Claim>(),
 					notBefore: token.NotBefore,
 					expires: token.ExpiresAt,
-					signingCredentials: securityToken.GetSigningCredentials()
+					signingCredentials: securityToken?.GetSigningCredentials()
```
j3parker, line 36: Behaviour change: previously we could throw `NullReferenceException` but now we would throw... something different

src/D2L.Security.OAuth2/Keys/JsonWebKey.cs
```diff
@@ -76,9 +73,17 @@ public static bool TryParseJsonWebKey(
...
 			try {
 				data = JsonSerializer.Deserialize<Dictionary<string, object>>( json );
+
+				if( data == null ) {
+					result = null;
+					error = "error deserializing JSON web key string";
+					exception = null;
+					useEncKey = false;
+					return false;
+				}
```
j3parker, line 86: Behaviour change but a good one

src/D2L.Security.OAuth2/Keys/JsonWebKey.cs
```diff
@@ -89,7 +94,7 @@ out bool useEncKey
 
 			if( data.TryGetValue( "use", out var use ) && use != null && use.ToString() != "sig" ) {
 				result = null;
-				error = "invalid 'use' value in JSON web key: " + data[ "use" ];
+				error = "invalid 'use' value in JSON web key: " + use;
```
j3parker, line 97: Missed this in a previous PR

src/D2L.Security.OAuth2/Keys/KeyManagementService.cs
```diff
@@ -35,18 +36,16 @@ OAuth2Configuration config
...
 			IPrivateKeyDataProvider privateKeys,
-			OAuth2Configuration config = null
+			OAuth2Configuration? config = null
 		) : this(
 			publicKeys,
 			privateKeys,
 			new DateTimeProvider(),
 			config ?? new OAuth2Configuration()
-		) {
-			config.CheckSanity();
```
j3parker, line 45: Moved the check to another constructor. Note that if `config` were `null` this actually would have crashed. Whoops!

src/D2L.Security.OAuth2/Keys/KeyManagementService.cs
```diff
@@ -60,11 +59,7 @@ await RefreshKeyAsync( now )
 				current = Volatile.Read( ref m_current );
 			}
 
-			if( current == null ) {
-				return null;
-			}
-
-			return current.Ref();
+			return current?.Ref();
```
j3parker, line 62: Not actually a behaviour change; I just rolled that `if` into this expression

src/D2L.Security.OAuth2/Principal/D2LPrincipal.cs
```diff
@@ -54,15 +54,17 @@ long ID2LPrincipal.UserId {
 			get {
 				AssertPrincipalTypeForClaim( PrincipalType.User, Constants.Claims.USER_ID );
 
-				return m_userId.Value;
+				// Bad API: NullReferenceException is expected
+				return m_userId!.Value;
```
j3parker, line 58: This I was less sure about making a behaviour change on without looking at usages, but I also noticed there are some tests for this behaviour so IMO for now "it is what it is"

src/D2L.Security.OAuth2/Provisioning/Default/AuthServiceClient.cs
```diff
@@ -108,12 +108,16 @@ IEnumerable<Scope> scopes
 					string errorMessage;
 
 					if( string.IsNullOrWhiteSpace( json ) ) {
-						errorMessage = response.ReasonPhrase;
+						errorMessage = response.ReasonPhrase ?? "unknown";
```
j3parker, line 111: Behaviour change, instead of NullReferenceException we will behave with a mid but consistent error message.

src/D2L.Security.OAuth2/Provisioning/Default/AuthServiceClient.cs
```diff
@@ -108,12 +108,16 @@ IEnumerable<Scope> scopes
...
-						errorMessage = response.ReasonPhrase;
+						errorMessage = response.ReasonPhrase ?? "unknown";
 					} else {
 						try {
 							var errorInfo = JsonSerializer.Deserialize<ErrorResponse>( json );
-							errorInfo.Validate();
-							errorMessage = string.Concat( errorInfo.Title, ": ", errorInfo.Detail );
+
+							if( errorInfo == null) {
+								errorMessage = "unknown error";
```
j3parker, line 117: Behaviour change but again I think a good one.

src/D2L.Security.OAuth2/Provisioning/Default/AuthServiceClient.cs
```diff
@@ -108,12 +108,16 @@ IEnumerable<Scope> scopes
 					string errorMessage;
 
 					if( string.IsNullOrWhiteSpace( json ) ) {
-						errorMessage = response.ReasonPhrase;
+						errorMessage = response.ReasonPhrase ?? "unknown";
 					} else {
 						try {
 							var errorInfo = JsonSerializer.Deserialize<ErrorResponse>( json );
-							errorInfo.Validate();
```
j3parker, line 115: Removed these `Validate()`s in favour of just requiring non-null in deserialization.
I was going to use a System.Text.Json feature that respects nullable annotations but it's only available in later versions of System.Text.Json and so I didn't want to create that version dependency just yet. So instead I did `[JsonRequired]`.

src/D2L.Security.OAuth2/Provisioning/Default/AuthServiceClient.cs
```diff
@@ -135,7 +139,14 @@ IEnumerable<Scope> scopes
...
 				try {
 					var grant = JsonSerializer.Deserialize<GrantResponse>( json );
-					grant.Validate();
+
+					if( grant == null) {
+						throw new AuthServiceException(
+							ServiceErrorType.ClientError,
+							message: "Couldn't deserialize response JSON"
+						);
+					}
```
j3parker, line 148: same sort of behaviour change for null deserialization

src/D2L.Security.OAuth2/Provisioning/Default/CachedAccessTokenProvider.cs
```diff
@@ -50,7 +50,7 @@ IEnumerable<Scope> scopes
 
 			if( cacheResponse.Success ) {
 				SecurityToken securityToken = m_tokenHandler.ReadToken( cacheResponse.Value );
-				if( securityToken.ValidTo > DateTime.UtcNow.Add( m_tokenRefreshGracePeriod ) ) {
+				if( securityToken.ValidTo > DateTime.UtcNow.Add( m_tokenRefreshGracePeriod ) && cacheResponse.Value != null ) {
```
j3parker, line 53: A behaviour change sort of but not in practice. The extra check makes the compiler happy and I don't want to wage war on that type right now.

src/D2L.Security.OAuth2/D2L.Security.OAuth2.csproj
```diff
@@ -1,44 +1,49 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
-	<PropertyGroup>
```
j3parker, line 2: Inconsistent whitespace in this file 😭

src/D2L.Security.OAuth2/D2L.Security.OAuth2.csproj
```diff
@@ -1,44 +1,49 @@
...
+    <IncludeSymbols>true</IncludeSymbols>
+  </PropertyGroup>
+
+  <PropertyGroup Condition="'$(TargetFramework)' != 'net472'">
+    <SuppressSyncGenerator>true</SuppressSyncGenerator>
+
+    <!-- nullable errors and annotations aren't enabled for .NET 4 because
+       the standard library doesn't really support that well. -->
+    <Nullable>enable</Nullable>
+    <WarningsAsErrors>nullable</WarningsAsErrors>
```
j3parker, line 19: I didn't enable this for TestFramework or tests in this PR because of the noise. Honestly the value, at this point, of turning it on in tests is not high.

omsmith: I feel like the value of actively skipping it for net472 is low though? The multi-targeting still gets you the better annotated checks for dotnet without the warnings for using nullable reference types without #nullable in Framework?

src/D2L.Security.OAuth2/Validation/AccessTokens/AccessTokenExtensions.cs
```diff
@@ -10,37 +9,39 @@ namespace D2L.Security.OAuth2.Validation.AccessTokens {
...
 		public static IEnumerable<Scope> GetScopes( this IAccessToken token ) {
-			string scopes = token.GetClaimValue( Constants.Claims.SCOPE );
+			var scopes = token.GetClaimValue( Constants.Claims.SCOPE );
 
 			if( string.IsNullOrEmpty( scopes ) ) {
 				return new Scope[] { };
 			}
 
-			Scope[] scopesArray = scopes
+			Scope?[] scopesArray = scopes
```
omsmith, line 31: Shouldn't be needed because of the filtering?

```suggestion
			Scope[] scopesArray = scopes
```

src/D2L.Security.OAuth2/Provisioning/Default/AuthServiceClient.cs
```diff
@@ -108,12 +108,16 @@ IEnumerable<Scope> scopes
...
-							errorInfo.Validate();
-							errorMessage = string.Concat( errorInfo.Title, ": ", errorInfo.Detail );
+
+							if( errorInfo == null) {
+								errorMessage = "unknown error";
+							} else {
+								errorMessage = string.Concat( errorInfo.Title, ": ", errorInfo.Detail );
+							}
 						} catch( Exception ) {
 							errorMessage = string.Concat( response.ReasonPhrase, ": ", json );
```
omsmith, line 122: Believe `response.ReasonPhrase` might be null here also?

